PR #8, opened 2015-02-04
Fix for android framerate issues

snmaynard:
This removes the initial stutter in framerate the first time notify is called.

I narrowed the issue down to the regex usage. I tried to pull out the regex into a static variable and then even tried using the compiled option to compile the regex ahead of time (which requires .net 2). I also tried just running the regex once in the register method. All of these options seemed to fail and this one was the only way i found to initialise the caches properly.

Can anyone think of a less hacky solution - or just something else to try?

### DIFF
--- a/src/Assets/Standard Assets/Bugsnag.cs
+++ b/src/Assets/Standard Assets/Bugsnag.cs
@@ -38,6 +38,7 @@ public class Bugsnag : MonoBehaviour {
             public static extern void ClearTab(string tabName);
         #elif UNITY_ANDROID && !UNITY_EDITOR
             public static AndroidJavaClass Bugsnag = new AndroidJavaClass("com.bugsnag.android.Bugsnag");
+            public static Regex unityExpression = new Regex ("(\\S+)\\s*\\(.*?\\)\\s*(?:(?:\\[.*\\]\\s*in\\s|\\(at\\s*\\s*)(.*):(\\d+))?", RegexOptions.IgnoreCase | RegexOptions.Multiline);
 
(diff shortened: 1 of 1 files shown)


src/Assets/Standard Assets/Bugsnag.cs
```diff
@@ -38,6 +38,7 @@ public class NativeBugsnag {
             public static extern void ClearTab(string tabName);
         #elif UNITY_ANDROID && !UNITY_EDITOR
             public static AndroidJavaClass Bugsnag = new AndroidJavaClass("com.bugsnag.android.Bugsnag");
+            public static Regex unityExpression = new Regex ("(\\S+)\\s*\\(.*?\\)\\s*(?:(?:\\[.*\\]\\s*in\\s|\\(at\\s*\\s*)(.*):(\\d+))?", RegexOptions.IgnoreCase | RegexOptions.Multiline);
```
loopj: When is this line called? When the app boots or when the first `Notify` is called?

snmaynard: Before any method on the class is called - in reality just before register.

loopj:
Maybe the JNI gets "booted up" in the first Notify call, that's all i can really see. Looks like the Regex should get compliled statically fine.

snmaynard:
var matches = regex.matches(stacktrace) is definitely the slow bit - i made the function do only that and it was slow the first time it was called

ConradIrwin:
:video_camera:

snmaynard:
Closing in favor of #9

PR #12, opened 2015-03-25
Allow passing context to notify

ConradIrwin:


src/Assets/Standard Assets/Bugsnag.cs
```diff
@@ -299,18 +300,28 @@ public enum LogSeverity {
...
     }
 
     public static void Notify(Exception e) {
+		if(e != null) {
+			var stackTrace = e.StackTrace;
+			if (stackTrace == null) {
+				stackTrace = new System.Diagnostics.StackTrace (1, true).ToString ();
+			}
+			
+			NotifySafely (e.GetType ().ToString (),e.Message, "warning", "", stackTrace);
```
snmaynard: tabs & strange indent

PR #24, opened 2016-06-24
Copies Android Unity Plugin

CodeHex:
The android notifier is now in aar format. This change fixes the rake script so that the correct android plugin is copied into the unity package.


kattrali:
@CodeHex thank you for the speedy turnaround!

PR #29, opened 2016-08-10
NotifyLevel changed to be static property

### LINKED ISSUE #28: NotifyLevel should be a static property
The docs indicate that the notify level should be changed via a static property on the Bugsnag class. https://docs.bugsnag.com/platforms/unity/#reporting-unhandled-errors

However, the property is only available as an instance property, requiring the user to find the Bugsnag object in the GameObject.

This should be changed to be a static property so it affects all Bugsnag objects and is easy to change.

CodeHex:
Allows users to set the notify level globally by making the NotifyLevel property static. Example project has also been updated

Note if anyone has set the notify level by setting it directly on the instance, this will be a breaking change.

Fixes #28


kattrali:
Hi @CodeHex, this makes a lot of sense. Thanks for patching this and the example so quickly :)

Issue #21, opened 2016-02-17
tvOS support


adrianmavenhut:
Hello.

The support for tvOS is on near-future roadmap?

Thanks.

kattrali:
It is, @adrianmavenhut! It will first need to be introduced downstream in bugsnag-cocoa, and there is an [issue you can watch about upcoming tvOS support](https://github.com/bugsnag/bugsnag-cocoa/issues/95).

PR #32, opened 2016-10-31
Support for tvOS and other fixes

### LINKED ISSUE #18: Specify available severity levels
Currently, in this notification plugin, severity is described as a string, which lead me to believe (erroneously) that Bugsnag supports any string as "severity" setting for an issue. However, the notifier API says that severity only has three valid values: "error", "warning" and "info". The best way to reflect it in C# is using an enum instead of string, and [dotNET notifier plugin already has one](https://github.com/bugsnag/bugsnag-dotnet/blob/6d59e4b80eeb3f3c4f3811aef26a14ca36761b05/src/Bugsnag/Enums.cs#L11).

Refactoring Unity notifier plugin to use enum instead of string internally would prevent such confusion.

### LINKED ISSUE #30: Compatibility with Unity 4.7
Loading up the example project in Unity 4.7, there seems to be some code that is not compiled in/out correctly. The new notifier code files should be checked that they run using both Unity 5 editor and Unity 4 editor. e.g. !UNITY_LT_5 doesn't work and should be replaced with UNITY_5_OR_NEWER.

### LINKED ISSUE #31: Use better folder structure
The folder structure used by bugsnag is quite messy and should be changed so the plugin resides in it's own sub folders, like most other plugins.

This makes it easier to create build systems and keep track of which files belong to Bugsnag.

Plugins
--> Android
-->-->Bugsnag

Bugsnag
--> Editor
-->--> BugsnagPostProcess
--> Busnag.cs

CodeHex:
Updates the Bugsnag cocoa notifier and enables tvOS support (#21). Also...
Fixes #31
Fixes #30
Fixes #18 

NOTE - The plugin files are not automatically being picked up by the Unity editor as tvOS and need to be activated manually. This can be done by navigating to Plugins > tvOS > Bugsnag, selecting all files and in the inspector selecting the tvOS checkbox.

We need to find a way to automatically do this, or add instructions on how to do this


CodeHex:
@kattrali I think this branch is pretty much done, any chance I could get some 👀

CodeHex:
The indentation was funny across all the files, so I've fixed it everywhere

PR #34, opened 2016-11-10
Init() to support runtime addition of Bugsnag component to existing GameObject

TimothyKLambert:
I created a public Init() method so the Bugsnag component can be added to an existing GameObject and initialized at runtime with a variable API key instead of predefined set through the Inspector.

We need this for our SDK because we support multiple API keys that can change for each project that uses our SDK so having to set the API key through the inspector before hand breaks our integration flow.


CodeHex:
Thanks for the contribution @TimothyKLambert, this looks good!

PR #36, opened 2016-11-14
Added custom mapping to severities

CodeHex:
Provides the ability to map Unity log levels to any Bugsnag severity.

Also changed default mapping to map Unity level Error and Unity level Assert to Bugsnag severity warning. Will require a major version bump as this will affect existing users severities if they changed the default NotifyLevel.


src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
```diff
@@ -285,6 +307,14 @@ public enum Severity {
         }
     }
 
+    public static void MapUnityLogToSeverity(LogSeverity unitySeverity, Severity bugsnagSeverity) {
+        if (customMapping.ContainsKey(unitySeverity)) {
+            customMapping[unitySeverity] = bugsnagSeverity;
+        } else {
+            customMapping.Add(unitySeverity, bugsnagSeverity);
+        }
```
CodeHex: Ah, for some reason I thought this wasn't supported in .NET 2.0, but it is! Changed!

snmaynard:
Will this allow people to do what they need to with mapping crashes through to error?

PR #38, opened 2016-11-21
Add log levels to Unity metadata

CodeHex:
When a unity log is handled and reported to Bugsnag, the log level is attached to the Unity metadata.


src/BugsnagUnity.mm
```diff
@@ -79,10 +79,18 @@ void Notify(char *errorClass, char *errorMessage, char *severity, char *context,
             ns_context = [Bugsnag configuration].context;
         }
 
+        // Indicate thats its a unity exception (with the received log level)
+        NSDictionary *unityData = nil;
+        if ([ns_logType isEqualToString: @""]) {
+            unityData = @{@"unityException": @true};
+        } else {
+            unityData = @{@"unityException": @true, @"unityLoglLevel": ns_logType};
```
TimothyKLambert: unityLogLevel - typo?

CodeHex: well spotted @TimothyKLambert! fixed typo

PR #41, opened 2017-04-19
[ENG-857] Add static API key

Pezzah:
A user has reported getting null pointer exceptions when manually creating the Bugsnag game object.

This seems to be because the API Key has not been set before the `awake` method gets called.

Adding a static way of setting the API Key should mean that the game object can be created manually and always have the API key set before construction.

### DIFF
--- a/example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
+++ b/example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
@@ -278,6 +278,7 @@ public class Bugsnag : MonoBehaviour {
     public static string[] SeverityValues = new string[]{"info", "error", "warning"};
 
     public string BugsnagApiKey = "";
+    public static string BugsnagApiKeyStatic = "";
     public bool AutoNotify = true;
 
     // Rate limiting section
@@ -368,6 +369,11 @@ public class Bugsnag : MonoBehaviour {
         maxCounts[unityLogType] = maxCount;
     }
 
+    // Used to set the API key when Bugsnag is being initialized
+    public static void SetApiKey(String apiKey) {
+        BugsnagApiKeyStatic = apiKey;
+    }
+
     string GetLevelName() {
 #if UNITY_5_OR_NEWER
       return SceneManager.GetActiveScene().name;
@@ -381,6 +387,13 @@ public class Bugsnag : MonoBehaviour {
         Init(null);
     }
 
+    public static Bugsnag createBugsnagInstance(GameObject gameObject, String ApiKey) {
+        Bugsnag.SetApiKey(ApiKey);
+        Bugsnag bugsnagInstance = gameObject.AddComponent<Bugsnag>();
+        bugsnagInstance.Init(); 
+        return bugsnagInstance;
+    }
+
     public void Init() {
         Init(null);
     }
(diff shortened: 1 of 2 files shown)


example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
```diff
@@ -378,10 +384,28 @@ public enum Severity {
...
-        Init(BugsnagApiKey);
+        Init(null);
+    }
+
+    public void Init() {
+		Init(null);
     }
 
-    public void Init(string apiKey)
+	public void Init(string apiKey)	{
```
CodeHex: Indentation is off

example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
```diff
@@ -378,10 +384,28 @@ public enum Severity {
...
+		// Try to work out which API key to use
+		if (!String.IsNullOrEmpty(apiKey)) {
+			InitInternal(apiKey);
+		} else if (!String.IsNullOrEmpty(BugsnagApiKey)) {
+			InitInternal(BugsnagApiKey);
+		} else if (!String.IsNullOrEmpty(BugsnagApiKeyStatic)) {
+			InitInternal(BugsnagApiKeyStatic);
+		} else {
+			// Will cause the native notifiers to throw an exception on Register
+			InitInternal(null);
```
CodeHex: Do we want to cause exceptions here?  Specifically on awake, maybe someone might want to initialise at a later stage. Maybe we shouldn't initialise it at all, so its similar to what would happen if you haven't called register in the client notifier. We could record an error log to indicate its not configured correctly.

example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
Pezzah: changed tabs to spaces

example/Assets/Standard Assets/Bugsnag/Bugsnag.cs
Pezzah: changed to log an error instead

Pezzah:
The manual setup for Bugsnag in unity will now look something like

```
GameObject bugsnagObject = new GameObject("Bugsnag");
GameObject.DontDestroyOnLoad(bugsnagObject);

Bugsnag.SetApiKey("your-api-key-here");
Bugsnag bugsnagInstance = bugsnagObject.AddComponent<Bugsnag>();
bugsnagInstance.Init(); 
```

snmaynard:
Why can't we do `bugsnagInstance.Init(apikey);`?

Pezzah:
Because by that point the object has been created and added to the component. The `Awake` method (https://github.com/bugsnag/bugsnag-unity/blob/e54400992e193f949e34e86b9b6ff05bf477b456/example/Assets/Standard%20Assets/Bugsnag/Bugsnag.cs#L385) can then get called before we get a chance to call `bugsnagInstance.Init()` manually.

I think that is what is happening to cause this problem

snmaynard:
Can we make awake check for null and do nothing if the apikey isnt set yet?

Pezzah:
yes, this change does that as well, but it feels like it would be better to allow the game to awake the objects when it wants to in case there are other side effects here

Pezzah:
Removing the static API key, the manual setup for Bugsnag in unity will now look something like

```
GameObject bugsnagObject = new GameObject("Bugsnag");
GameObject.DontDestroyOnLoad(bugsnagObject);

Bugsnag bugsnagInstance = bugsnagObject.AddComponent<Bugsnag>();
bugsnagInstance.Init("your-api-key-here"); 
```

If the `awake` method gets called early, then there will be an error logged, but Bugsnag will still start to work once the `Init()` method is called manually

Pezzah:
Added a create method to make the creation nicer, so the new way to create manually will be
```
GameObject bugsnagObject = new GameObject("Bugsnag");
GameObject.DontDestroyOnLoad(bugsnagObject);
Bugsnag.createBugsnagInstance(bugsnagObject, "your-api-key-here"); 
```

PR #44, opened 2017-06-12
fix require 'shellwords' escaping

wrossmck:
`require 'shellwords'` => `require \\\"shellwords\\\"`

On some OSX El Capitan instances I found that the `'` would break string escaping. Using a standardised escaping system like the rest of the string fixes this escaping issue across my build farm.


kattrali:
Works perfectly on El Cap and newer versions. :)

PR #46, opened 2017-07-14
Load bugsnag-js as a dependency

kattrali:
In the case that Bugsnag fails to load, either due to networking issues
or content blockers, _load it from disk to avoid this entirely!_ Added 
bugsnag-js as a submodule dependency, like iOS/Android/Mac.

~print a warning message and mock the global Bugsnag
object to avoid further errors. It also provides a debuggable interface
for using Bugsnag from JS to set config options.~

Fixes #43

Example to reproduce:
* Rebuild the package with `rake build`
* Open a project in unity, then `Bugsnag.unitypackage`
* Install uBlock Origin in Firefox
* Generate a WebGL build of an app including Bugsnag
* Confirm that everything is fine.

PR #51, opened 2017-08-24
Upgrade dependencies, fix notifier details

### LINKED ISSUE #50: v3.3.2 Bugsnag.unitypackage has compile error with Unity 5.5
At Bugsnag.cs line 431, `Application.identifier` is used instead of `Application.bundleIdentifier`.
Maybe Automatic API Updater ran accidentally when exporting Bugsnag.unitypackage.

kattrali:
This changeset allows us to correctly report the notifier name and version from unity, as well as adding an extension point for customizations on Android.

## Changes
* Update `bugsnag-android` to 3.9.0. We can move to 4.0 in a more comprehensive release
* Add a new tiny library, `bugsnag-android-unity` as an extension point to customize and deliver notifications. All log- and manual `Notify()`- generated error reports now run through a callback for tweaking a few fields, as well as reducing the JNI calls required.
* Fix #50 by gating the `Application.identifier` change
* Update release instructions, including where to change the notifier version

## Notes
* The version number currently exists in two places. We should refine this eventually to reduce the risk of sending different version numbers from different platforms.
* WebGL sends via the JS endpoint and also uses JS notifier details. For now, this seems reasonable, as changing it may have _implications._

### DIFF
--- a/bugsnag-android-unity/build.gradle
+++ b/bugsnag-android-unity/build.gradle
@@ -1,7 +1,6 @@
 buildscript {
     repositories {
         jcenter()
-        mavenCentral()
     }
 
     dependencies {
--- a/bugsnag-android-unity/build.gradle
+++ b/bugsnag-android-unity/build.gradle
@@ -1,0 +1,25 @@
+buildscript {
+    repositories {
+        jcenter()
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:2.3.1'
+    }
+}
+
+apply plugin: 'com.android.library'
+android {
+    compileSdkVersion Integer.parseInt(project.ANDROID_BUILD_SDK_VERSION)
+    buildToolsVersion project.ANDROID_BUILD_TOOLS_VERSION
+
+    defaultConfig {
+        minSdkVersion Integer.parseInt(project.ANDROID_BUILD_MIN_SDK_VERSION)
+        targetSdkVersion Integer.parseInt(project.ANDROID_BUILD_TARGET_SDK_VERSION)
+    }
+}
+
+dependencies {
+    compile project(':bugsnag-android')
+}
--- a/bugsnag-android-unity/gradle.properties
+++ b/bugsnag-android-unity/gradle.properties
@@ -1,0 +1,12 @@
+GROUP=com.bugsnag
+
+POM_NAME=Bugsnag Android Unity
+POM_DESCRIPTION=Official Unity extension for Bugsnag on Android
+POM_URL=https://bugsnag.com
+POM_ARTIFACT_ID=bugsnag-android-unity
+POM_PACKAGING=aar
+
+ANDROID_BUILD_TARGET_SDK_VERSION=21
+ANDROID_BUILD_TOOLS_VERSION=25.0.2
+ANDROID_BUILD_SDK_VERSION=25
+ANDROID_BUILD_MIN_SDK_VERSION=4
--- a/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
+++ b/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
@@ -4,13 +4,7 @@ import java.lang.String;
 
 import android.content.Context;
 
-import com.bugsnag.android.Bugsnag;
-import com.bugsnag.android.Callback;
-import com.bugsnag.android.Client;
-import com.bugsnag.android.Error;
-import com.bugsnag.android.MetaData;
-import com.bugsnag.android.Report;
-import com.bugsnag.android.Severity;
+import com.bugsnag.android.*;
 
 public class UnityClient {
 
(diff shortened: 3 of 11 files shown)


bugsnag-android-unity/build.gradle
```diff
@@ -0,0 +1,25 @@
+buildscript {
+    repositories {
+        jcenter()
+        mavenCentral()
```
fractalwrench: Can get rid of mavenCentral as jcenter supplies the same dependencies and is faster

bugsnag-android-unity/gradle.properties
```diff
@@ -0,0 +1,12 @@
...
+POM_NAME=Bugsnag Android Unity
+POM_DESCRIPTION=Official Unity extension for Bugsnag on Android
+POM_URL=https://bugsnag.com
+POM_ARTIFACT_ID=bugsnag-android-unity
+POM_PACKAGING=aar
+
+ANDROID_BUILD_TARGET_SDK_VERSION=21
+ANDROID_BUILD_TOOLS_VERSION=25.0.2
+ANDROID_BUILD_SDK_VERSION=25
+ANDROID_BUILD_MIN_SDK_VERSION=4
```
fractalwrench: Did we want to increase this to 14, or leave that until we've updated to Bugsnag Android 4.0.0?

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -0,0 +1,56 @@
...
+    public static void init(Context androidContext, String apiKey) {
+        Bugsnag.init(androidContext, apiKey);
+    }
+
+    public static void notify(String name, String message, String context, StackTraceElement[] stacktrace, Severity severity, String logLevel) {
+        Bugsnag.notify(name, message, stacktrace, new UnityCallback(context, severity, logLevel));
+    }
+}
+
+class UnityCallback implements Callback {
```
fractalwrench: Could this top-level class be moved to a separate source file?

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -0,0 +1,56 @@
...
+    static final String NOTIFIER_NAME = "Bugsnag Unity (Android)";
+    static final String NOTIFIER_VERSION = "3.4.0";
+    static final String NOTIFIER_URL = "https://github.com/bugsnag/bugsnag-unity";
+
+    final private Severity severity;
+    final private String context;
+    final private String logLevel;
```
fractalwrench: Great that it's final. The Java convention is generally to have the access modifier first i.e. `private final`
https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.3.1

fractalwrench:
A few other things:

- Opening the example project prompted me to update the API (I'm running Unity 2017.1.of3)
- Inspecting the Bugsnag GameObject in the hierarchy says that the script is missing, meaning the API key can't be changed. Adding the location of the script manually makes it show up.
- WebGL didn't send errors to test project with correct API Key
- Dashboard errors in iOS always showed "Unknown file" in the stacktrace
- Blank blue screen displayed on Android

kattrali:
> * Opening the example project prompted me to update the API (I'm running Unity 2017.1.of3)

I don't see how this can be fixed. If we update the API, the library will break for users on older versions of unity (e.g. #50). I've gated the code in an `#if` for unity versions, though I don't understand why the editor doesn't respect that.

bugsnag-android-unity/gradle.properties
kattrali: I made a note to move the version when we upgrade to 4.0.0

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
kattrali: Yeah 🤔 There's no reason the two classes need to be in the same file.

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
kattrali: Ack. Good catch!

kattrali:
> * WebGL didn't send errors to test project with correct API Key

Hmm, that is odd. Nothing has changed in there since #46 🤔 

> * Dashboard errors in iOS always showed "Unknown file" in the stacktrace

I think This is Normal™

> * Blank blue screen displayed on Android

Based on prior discussions, it sounds like a layout issue. If this works with another project, I think we're clear to move on it.

kattrali:
@fractalwrench I rebased to pull in the new example project; however you'd need to `rake build; open Bugsnag.unitypackage` again to retest (I haven't checked in the example project changes to keep the diff cleaner).

fractalwrench:
@kattrali that command appears to fail on a fresh checkout following the Contributing.md instructions:
`git clone --recursive git@github.com:bugsnag/bugsnag-unity`
`rake build; open Bugsnag.unitypackage`

> A problem occurred configuring project ':example'.
> > Could not resolve all dependencies for configuration ':example:classpath'.
>    > Could not find com.android.tools.build:gradle:3.0.0-beta3.
>      Searched in the following locations:
>          https://repo1.maven.org/maven2/com/android/tools/build/gradle/3.0.0-beta3/gradle-3.0.0-beta3.pom
>          https://repo1.maven.org/maven2/com/android/tools/build/gradle/3.0.0-beta3/gradle-3.0.0-beta3.jar
>          file:/Users/jamielynch/.m2/repository/com/android/tools/build/gradle/3.0.0-beta3/gradle-3.0.0-beta3.pom
>          file:/Users/jamielynch/.m2/repository/com/android/tools/build/gradle/3.0.0-beta3/gradle-3.0.0-beta3.jar
>      Required by:
>          bugsnag-android:example:unspecified > com.bugsnag:bugsnag-android-gradle-plugin:3.0.0-beta

Is there any additional setup I need to do with the gitmodules to test this?

kattrali:
Yes - when switching from the master branch, you'll need to run `git submodule update` to get the correct submodule commits on this branch.

kattrali:
That probably needs to be recursive actually: `git submodules update --recursive`

PR #63, opened 2017-10-12
Synchronize android build info into unity module

kattrali:
Set a symbolic link to use the same gradle.properties file in bugsnag-android-unity as on bugsnag-android. Given that the POM metadata is unused, the duplicated artifact metadata has no effect.


fractalwrench:
@kattrali we'd also need some sort of solution for `project.ext` variables if we went with this approach: https://github.com/bugsnag/bugsnag-android/blob/master/build.gradle

I'm currently working on a ticket for this one - should I hold off?

kattrali:
Those don't necessarily need to be the same, do they? For instance, if I added tests with a different mockito version dependency than what bugsnag-android uses, is that a problem?

fractalwrench:
True, they don't necessarily need to be the same.

An issue with the symlink approach is that the Android repo has conflicting/missing properties compared to this repo, e.g. `POM_ARTIFACT_ID`: https://github.com/bugsnag/bugsnag-android/blob/master/gradle.properties

kattrali:
True but they don't matter for the unity library, and the bundle is still generated normally

PR #64, opened 2017-10-12
Use android unity module

fractalwrench:


fractalwrench:
Note: `xcpretty` appears to be broken on High Sierra, and the `Rakefile` would need to rename `unity-release.aar` to the previous aar name.

PR #65, opened 2017-10-22
Add native-Unity BugSnag support.

zorbathut:
This probably does some things slightly wrong regarding your conventions, and note that I may have accidentally broken Android support; you'll want to test that (I don't have the ability to.)

On the other hand, seems to work pretty well for my use case, so, if it's useful for you, have fun!


fractalwrench:
Thanks for the PR @zorbathut! We'll try and get round to reviewing it soon.

martin308:
@zorbathut thanks for your contribution! Sorry it has taken us so long to get back to this. We are actively working on this project now and will have something to show in the near future

PR #67, opened 2017-10-30
Allow multiline message

shinjiyoshida:
When an exception with multiline message was captured by Application.logMessageReceived, its lines was omitted except the first line because default Regex ctor is used.


fractalwrench:
Thanks for the PR @shinjiyoshida - we'll try and test this out soon. Just to clarify - by multiline message, does that mean a String containing newlines?

shinjiyoshida:
Yes!

PR #73, opened 2018-05-02
fix(android): Cache Client object for reuse

kattrali:
Passes all usage of the android client to the unity wrapper, avoiding calls to `Bugsnag.getClient()` and preventing multiple clients from being registered (unintentionally or otherwise) from the unity layer.

### DIFF
--- a/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
+++ b/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
@@ -10,16 +10,24 @@ import com.bugsnag.android.*;
 
 public class UnityClient {
 
+    static Client client;
+
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
+        if (client != null) {
+            client.disableExceptionHandler();
+        }
         Configuration config = new Configuration(apiKey);
         config.setAutoCaptureSessions(trackSessions);
-        Bugsnag.init(androidContext, config);
+        client = new Client(androidContext, config);
     }
 
     public static void notify(String name, String message,
                               String context, StackTraceElement[] stacktrace,
                               Severity severity, String logLevel,
                               String severityReason) {
+        if (client == null) {
+            return;
+        }
         Throwable t = new BugsnagException(name, message, stacktrace);
 
         Map<String, Object> data = new HashMap<>();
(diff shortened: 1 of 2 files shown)


bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -10,32 +10,115 @@
...
+    static Client client;
+
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
+        if (client != null) {
+            client.disableExceptionHandler();
+        }
         Configuration config = new Configuration(apiKey);
         config.setAutoCaptureSessions(trackSessions);
-        Bugsnag.init(androidContext, config);
+        client = new Client(androidContext, config);
```
fractalwrench: I'm not sure that this method is thread-safe - `init` could be called concurrently multiple times from Unity, and if the `client` field hasn't been assigned, a client could be constructed twice or more.

fractalwrench: N.B. bugsnag-android uses the same pattern so any change should also be applied there

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -10,32 +10,115 @@
 
 public class UnityClient {
 
+    static Client client;
+
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
+        if (client != null) {
+            client.disableExceptionHandler();
```
snmaynard: My inclination would be to no op the method if client is non null. We could use an atomicboolean to ensure the method is only run once to also solve @fractalwrench's comment below? I think this will also mean that the config changes persist properly. Only downside as @martin308 says is that any changes to apikey or tracksessions by calling init multiple times would not be possible - but personally that seems both unlikely and a reasonable api restriction.

Thoughts?

fractalwrench: @martin308 I think a couple of the above points are also applicable to the `Register` method in `Bugsnag.cs`. I'm not too familiar with how C# handles concurrency - would you be able to have a look and see whether there are any potential issues?

src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
```diff
@@ -174,61 +174,57 @@ public class NativeBugsnag {
...
         public static void SetReleaseStage(string releaseStage) {
             if (!CheckRegistration()) return;
-            Bugsnag.CallStatic ("setReleaseStage", releaseStage);
+            BugsnagUnity.CallStatic ("setReleaseStage", releaseStage);
         }
 
         public static void SetNotifyReleaseStages(string releaseStages) {
             if (!CheckRegistration()) return;
-            Bugsnag.CallStatic ("setNotifyReleaseStages", releaseStages.Split (','));
+            BugsnagUnity.CallStatic ("setNotifyReleaseStages", releaseStages.Split (','));
```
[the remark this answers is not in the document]
martin308: It looks like it wouldn't but it would have the same functionality as it currently does in regards to null values

src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
```diff
@@ -174,61 +174,57 @@ public class NativeBugsnag {
...
-            } else {
-                Bugsnag.CallStatic ("disableExceptionHandler");
-            }
+            BugsnagUnity.CallStatic ("setAutoNotify", autoNotify);
         }
 
         public static void SetContext(string context) {
             if (!CheckRegistration()) return;
-            Bugsnag.CallStatic ("setContext", context);
+            BugsnagUnity.CallStatic ("setContext", context);
```
bullmo: I may be misunderstanding how this is configured but in other places in this file we are calling `Bugsnag` directly rather than through `NativeBugsnag` e.g. [setContext](https://github.com/bugsnag/bugsnag-unity/blob/d8dc980b82a6006e5554f8a652b85f9076c7b42a/src/Assets/Standard%20Assets/Bugsnag/Bugsnag.cs#L478).

Should we only be configuring Bugsnag through the wrapper?

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -10,32 +10,119 @@
...
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
-        Configuration config = new Configuration(apiKey);
-        config.setAutoCaptureSessions(trackSessions);
-        Bugsnag.init(androidContext, config);
+        if (client == null) {
+            synchronized (lock) {
+                if (client == null) {
+                    Configuration config = new Configuration(apiKey);
+                    config.setAutoCaptureSessions(trackSessions);
+                    client = new Client(androidContext, config);
```
bullmo: How come this is using `new Client` rather than `Bugsnag.init` (which returns a `Client`)?

fractalwrench:
I had a quick look at the C# code and it appears as though `init` can be called twice in certain circumstances. For example, a user might call [`createBugsnagInstance`](https://github.com/bugsnag/bugsnag-unity/blob/d8dc980b82a6006e5554f8a652b85f9076c7b42a/src/Assets/Standard%20Assets/Bugsnag/Bugsnag.cs#L443) which invokes `init`, then I [believe](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Awake.html) that `Awake` may also invoke `init`.

Sorry for just picking this up at this stage of the review - but is this likely to cause an issue with initialisation, and if so, how could we prevent it? I'm not sure I fully understand exactly what all the entrypoints are doing (I count 5 - init, the static factory methods, and adding the script as a component), so it'd be good to consider whether there are other issues in this area.

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
martin308: good catch! Looks like it changed here https://github.com/bugsnag/bugsnag-unity/pull/73/commits/27e32c4939598ef668fc95781f841aec352e8914#r185431610, @kattrali do you know the reasoning?

src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
martin308: It is quite confusing and I've spent quite a bit of time trying to pick through the flow of everything.

It looks like all of the methods on `Bugsnag` pass through to `NativeBugsnag` after performing some validation on the parameters. 

For Android the flow should go `Bugsnag` -> `NativeBugsnag` -> `BugsnagUnity` (the java class). It gets more confusing though, as in the implementation of `NativeBugsnag` for Android there is a static field named `Bugsnag` which is the java class for the Android client. I believe that this could be removed to reduce the confusion.

fractalwrench:
Approving as after discussing with Martin, this area will be revamped imminently in future work.

PR #76, opened 2018-05-14
Cache Client object for reuse - preserve configuration

martin308:
If a new client is ever created then the user will lose any previously set configuration values. This attempts to preserve the values that we can from the previous configuration object

### DIFF
--- a/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
+++ b/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
@@ -19,7 +19,6 @@ public class UnityClient {
         }
         Configuration config = new Configuration(apiKey);
         config.setAutoCaptureSessions(trackSessions);
-        client = new Client(androidContext, config);
         if (configuration != null) {
           // preserve the configuration values that we can that were set on the
           // previous clients configuration. We can't reuse the configuration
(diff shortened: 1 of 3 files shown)


bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -19,6 +20,18 @@ public static void init(Context androidContext, String apiKey, boolean trackSess
         Configuration config = new Configuration(apiKey);
         config.setAutoCaptureSessions(trackSessions);
         client = new Client(androidContext, config);
```
kattrali: Possibly small issue: client initialization should be after setting all of the config properties in case of unexpected (background?) interactions between enabling session tracking and setting endpoints

martin308: good point, changed

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -11,14 +11,27 @@
 public class UnityClient {
 
     static Client client;
+    static Configuration configuration;
```
fractalwrench: A `volatile` modifier will ensure that an up-to-date value of this variable can be seen on any thread.

fractalwrench:
I believe this PR shouldn't be required due to my suggestion in #73 - if we decide not to go with this for whatever reason, then this looks like a reasonable approach.

martin308:
closing this in favour of #77

PR #77, opened 2018-05-15
Cache Client object for reuse - no-op on subsequent calls

martin308:
Here is the alternate approach suggested by both @snmaynard and @fractalwrench where we only create the client once. Any additional calls to `init` will be a no-op

### DIFF
--- a/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
+++ b/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
@@ -10,15 +10,19 @@ import com.bugsnag.android.*;
 
 public class UnityClient {
 
-    static Client client;
+    static volatile Client client;
+    private static Object lock = new Object();
 
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
-        if (client != null) {
-            client.disableExceptionHandler();
+        if (client == null) {
+            synchronized (lock) {
+                if (client == null) {
+                    Configuration config = new Configuration(apiKey);
+                    config.setAutoCaptureSessions(trackSessions);
+                    client = new Client(androidContext, config);
+                }
+            }
         }
-        Configuration config = new Configuration(apiKey);
-        config.setAutoCaptureSessions(trackSessions);
-        client = new Client(androidContext, config);
     }
 
     public static void notify(String name, String message,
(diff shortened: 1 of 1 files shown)


bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -10,15 +10,19 @@
...
 public class UnityClient {
 
-    static Client client;
+    static volatile Client client;
+    private static Object lock = new Object();
 
     public static void init(Context androidContext, String apiKey, boolean trackSessions) {
-        if (client != null) {
-            client.disableExceptionHandler();
+        if (client == null) {
```
fractalwrench: This check is already carried out in the `synchronized` block so could be removed.

fractalwrench:
The Java approach looks good - I left a couple of comments inline but otherwise I think that's ready to go.

From the slack conversation one of the concerns about this approach were that it doesn't take into account `init` being called with different parameters later on, which could lead to unexpected behaviour. Personally I think that it's reasonable to expect for the API key to only be set once at initialisation, and for `init` to be called only once. To reduce the risk of confusion, maybe it would be worth bumping to a minor version, and adding a brief warning to the changelog about this alteration in behaviour?

@kattrali can you give secondary review since this touches concurrency?

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
martin308: This avoids taking the 'expensive' lock when we don't need to

bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/UnityClient.java
```diff
@@ -10,15 +10,19 @@
 
 public class UnityClient {
 
-    static Client client;
+    static volatile Client client;
+    private static Object lock = new Object();
```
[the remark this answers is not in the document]
martin308: Synchronizing on a public class would seem unwise. Anyone can also take a lock on `UnityClient` which would be bad news.

From my reading it seems that `final` keywords effects on fields is to mark them as 'assign' once. How would this effect its nullability? I'll add it though as that is the behavior that we want 😄

martin308:
Going to merge this as it seems the consensus is to go down this path, we can do a final review in #73

PR #80, opened 2018-06-29
CI Build

martin308:
This will build the current unitypackage.

We should also configure travis to create the GitHub release and upload the built package. Just need to determine which accounts credentials should be used to do this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os: osx
 osx_image: xcode9.3
 env:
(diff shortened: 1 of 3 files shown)


martin308:
@fractalwrench added caching for bundler/gradle/android following the patterns in bugsnag-android. I added caching of both the mono and unity downloads but it was much slower than downloading them from their original sources so I removed it but made them download in parallel which sped it up slightly.

PR #85, opened 2018-08-13
Add windows device and app data to match native notifiers

martin308:
Include additional information for the device and app tab on windows machines. In order to make this easier the client objects where restructured so to use composition instead of inheritance. This made it easier to include additional custom behavior for the windows client. 

Additionally I discovered that some of the JNI code that we were using does not work in the latest version of unity so I had to revert back to the manual JNI code


src/Assets/Standard Assets/Bugsnag/BugsnagBehaviour.cs
```diff
@@ -50,10 +50,12 @@ void Awake()
...
-      if (hasFocus && Bugsnag.Client.Configuration.AutoCaptureSessions)
-      {
-        Bugsnag.Client.SessionTracking.StartSession();
-      }
+      Bugsnag.Client.SetApplicationState(hasFocus);
+    }
+
+    void OnApplicationPause(bool paused)
+    {
+      Bugsnag.Client.SetApplicationState(!paused);
```
fractalwrench: This is very minor but the variable naming/method signature here reads as though the application is being set as not paused (even though the opposite is the case)

src/BugsnagUnity/Clients/Client.cs
```diff
@@ -0,0 +1,223 @@
...
+    public void SetApplicationState(bool inFocus)
+    {
+      if (inFocus)
+      {
+        Stopwatch.Start();
+
+        if (Configuration.AutoCaptureSessions)
+        {
+          SessionTracking.StartSession();
+        }
```
fractalwrench: Is there any possibility that the Unity callbacks i.e. `OnApplicationFocus` could be called multiple times? If so, this could lead to unwanted sessions.

src/BugsnagUnity/Clients/AndroidClient.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        // the bugsnag-android notifier uses Activity lifecycle tracking to
+        // determine if the application is in the foreground. As the unity
+        // activity has already started at this point we need to tell the
+        // notifier about the activity and the fact that it has started.
+        using (var sessionTracker = JavaClient.Get<AndroidJavaObject>("sessionTracker"))
+        using (var activityClass = activity.Call<AndroidJavaObject>("getClass"))
+        {
+          var activityName = activityClass.Call<string>("getSimpleName");
+          sessionTracker.Call("updateForegroundTracker", activityName, true, 0L);
```
fractalwrench: This needs to be set to the Java time in ms, otherwise the created session won't be used in the Releases Dashboard.

src/BugsnagUnity/Payload/Device.cs
```diff
@@ -23,6 +21,13 @@ internal Device(string hostname)
       this.AddToPayload("timezone", TimeZone.CurrentTimeZone.StandardName);
       this.AddToPayload("osName", OsName);
       this.AddToPayload("time", DateTime.UtcNow);
+
+      var matches = Regex.Match(Environment.OSVersion.VersionString, "\\A(?<osName>[a-zA-Z ]*) (?<osVersion>[\\d\\.]*)\\z");
```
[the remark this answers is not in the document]
martin308: I will add a comment about what I expect the string to look like. I don't have access to multiple versions of windows so I can only go off what I have seen for the format of the string. Above this line we set `osName` to the full string so this is the fallback

src/BugsnagUnity/Clients/MacOsClient.cs
```diff
@@ -0,0 +1,114 @@
...
+    [DllImport("bugsnag-osx", EntryPoint = "bugsnag_setMetadata")]
+    static extern void AddMetadata(IntPtr configuration, string tab, string[] metadata, int metadataCount);
+
+    [DllImport("bugsnag-osx", EntryPoint = "bugsnag_retrieveDeviceData")]
+    static extern void RetrieveDeviceData(IntPtr instance, Action<IntPtr, string, string> populate);
+
+    [DllImport("bugsnag-osx", EntryPoint = "bugsnag_retrieveAppData")]
+    static extern void RetrieveAppData(IntPtr instance, Action<IntPtr, string, string> populate);
+
+    MacOSConfiguration Configuration { get; }
```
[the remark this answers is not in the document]
martin308: Good spot!

src/BugsnagUnity/Clients/AndroidClient.cs
martin308: Im not sure what you mean by the Java time in ms. The branch of code this executes only does this `foregroundActivities.add(activityName);` so that parameter is unused.

src/BugsnagUnity/Payload/Device.cs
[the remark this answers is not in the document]
fractalwrench: That sounds reasonable!

src/BugsnagUnity/Clients/AndroidClient.cs
fractalwrench: Passing in `0L` will result in no session being started within [updateForegroundTracker](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/SessionTracker.java#L279). If that's intentional and a session is started via some other method then this is fine, otherwise we would need to pass in the current Date in milliseconds since the Java epoch.

martin308: I only want this line of code to run `foregroundActivities.add(activityName);` session tracking is disabled in the android notifier as we do it in the c# code

src/Assets/Standard Assets/Bugsnag/BugsnagBehaviour.cs
```diff
@@ -50,10 +50,13 @@ void Awake()
...
-      {
-        Bugsnag.Client.SessionTracking.StartSession();
-      }
+      Bugsnag.Client.SetApplicationState(hasFocus);
+    }
+
+    void OnApplicationPause(bool paused)
+    {
+      var hasFocus = !paused;
+      Bugsnag.Client.SetApplicationState(hasFocus);
```
martin308: @fractalwrench what do you think of this code now, is the intent clearer?

src/BugsnagUnity/Clients/MacOsClient.cs
```diff
@@ -0,0 +1,114 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using AOT;
+using BugsnagUnity.Payload;
+
+namespace BugsnagUnity
+{
+  class MacOsClient : INativeClient
```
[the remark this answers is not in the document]
martin308: I found a reasonable pattern and implemented it!

src/BugsnagUnity/Clients/Client.cs
martin308: It could indeed be called multiple times, each time though I believe we would want a new session

src/Assets/Standard Assets/Bugsnag/BugsnagBehaviour.cs
fractalwrench: lgtm!

PR #86, opened 2018-08-14
Implement basic maze-runner test

martin308:
Adds an extremely simple maze-runner test that creates a Mac OS X game that does a manual notify and ensures that we receive it and it looks basically correct.

We can add additional things on top of this if we think it's a good start.


Rakefile
```diff
@@ -165,6 +165,10 @@ namespace :plugin do
     rm_f package_output
     unity "-projectpath", project_path, "-exportpackage", "Assets", package_output
   end
+
+  task maze_runner: %w[plugin:export] do
+    sh "bundle", "exec", "bugsnag-maze-runner", "--verbose"
```
fractalwrench: `--verbose` has added a surprising amount of time onto the build times for other platforms, when we've finished debugging the CI setup it would be worth removing this.

src/BugsnagUnity/IDelivery.cs
```diff
@@ -61,7 +61,7 @@ IEnumerator PushToServer(IPayload payload, byte[] body)
     {
       using (var req = new UnityWebRequest(payload.Endpoint.ToString()))
       {
-        req.SetRequestHeader("Content-Type", "application/json; charset=utf-8");
+        req.SetRequestHeader("Content-Type", "application/json");
```
fractalwrench: do we not need to set the charset anymore? If this fails a mazerunner step, we should update the mazerunner assertion

features/scripts/create_unity_project.sh
```diff
@@ -0,0 +1,17 @@
...
+pushd "${0%/*}"
+  pushd ../..
+    package_path=`pwd`
+  popd
+  pushd ../fixtures
+    rm -rf unity_project
+    Unity -nographics -quit -batchmode -logFile unity.log -createProject unity_project
+    Unity -nographics -quit -batchmode -logFile unity.log -projectpath unity_project -importPackage "$package_path/Bugsnag.unitypackage"
+    cp Main.cs unity_project/Assets/Main.cs
+    BUGSNAG_APIKEY=a35a2a72bd230ac0aa0f52715bbdc6aa Unity -nographics -quit -batchmode -logFile unity.log -projectpath unity_project -executeMethod "Main.CreateScene"
```
fractalwrench: The API key is set within the scenario, so shouldn't be required here

features/steps/unity_steps.rb
```diff
@@ -0,0 +1,17 @@
...
+  ])
+end
+
+When("run the application") do
+  run_required_commands([
+    ["features/scripts/launch_unity_application.sh"]
+  ])
+end
+
+Given("I configure the bugsnag endpoint") do
```
fractalwrench: May want to consider renaming this to 'bugsnag notify endpoint' for when we add sessions

fractalwrench:
This looks like a reasonable approach - added a few minor nits around naming etc. The build does appear to be failing on Travis CI

Rakefile
martin308: yeh the build is failing so I'm trying to find out why

src/BugsnagUnity/IDelivery.cs
martin308: It is indeed due to maze-runner, I don't think it's necessary though as JSON is by default (possibly only?) in utf8

PR #87, opened 2018-08-21
Stacktrace improvements

martin308:
In the case where we have been given an exception with no stack trace (this is usually the case when we are given an exception that has not been thrown) we previously tried to generate a stack trace using the current stack trace. This does not work in Unity inside of the generator code. 

The fix here is to generate a stack trace outside of the generator code that we can use in the case where we are not able to get a stacktfrace from the provided exception.

### DIFF
--- a/src/BugsnagUnity/Payload/StackTraceLine.cs
+++ b/src/BugsnagUnity/Payload/StackTraceLine.cs
@@ -18,8 +18,6 @@ namespace BugsnagUnity.Payload
 
     StackFrame[] AlternativeStackTrace { get; }
 
-    string BoundaryMethod { get; }
-
     private static string[] StringSplit { get; } = { Environment.NewLine };
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/Payload/StackTraceLine.cs
```diff
@@ -92,7 +92,7 @@ public IEnumerator<StackTraceLine> GetEnumerator()
         {
           // if the exception has not come from a stack trace then we need to
           // skip the frames that originate from inside the notifier code base
-          var currentStackFrameIsNotify = stackFrame.MethodName.StartsWith("Bugsnag.Client.Notify");
+          var currentStackFrameIsNotify = stackFrame.MethodName.StartsWith("BugsnagUnity.Client.Notify", StringComparison.InvariantCulture);
```
fractalwrench: Is there a way to sanitise the stacktrace before passing the value into `StackTraceLine`? Ideally this class shouldn't require knowledge about internal method names elsewhere in the codebase

PR #90, opened 2018-08-22
Move Event getters/setters into Report

martin308:
1. Make Report.Event non-public (this is the only place the Event class is used , so the class should be non-public as well)
1. Move getters/setters for User, Severity, HandledState (private), GroupingHash, Exceptions, Device, Context, Breadcrumbs, App, IsHandled, LogType, and MetaData from Event to Report.

### DIFF
--- a/src/BugsnagUnity/Payload/Report.cs
+++ b/src/BugsnagUnity/Payload/Report.cs
@@ -6,12 +6,30 @@ namespace BugsnagUnity.Payload
 {
   public class Report : Dictionary<string, object>, IPayload
   {
+    /// <summary>
+    /// Gets the configuration used to construct this report.
+    /// </summary>
+    /// <value>The configuration.</value>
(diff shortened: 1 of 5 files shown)


src/BugsnagUnity/Payload/Report.cs
```diff
@@ -27,11 +28,51 @@ internal Report(IConfiguration configuration, Event @event)
...
+    Event Event { get; }
+
+    internal HandledState OriginalSeverity => Event.OriginalSeverity;
 
     public void Ignore()
     {
       Ignored = true;
     }
+
+    public Metadata Metadata => Event.Metadata;
```
fractalwrench: Do we need inline docs for any of these properties, or is that being addressed separately?

martin308: Good idea, I'll add them there as well. I also need to update the docs site which I have a note to do

Issue #88, opened 2018-08-21
Android Crash on start 


willispinaud:
### Description
Trying to install the alpha version on Bugsnag on our Unity project

### Issue
When starting the application it crashes

### Environment

* bugsnag-unity version: 4.0.0 alpha 1
* Unity version: 2017.4.9f1
* Operating system name and version: MacOs 10.13.5
* Target platform names and versions: Android 6.0.1
* Initializing bugsnag via the Unity UI or in code? Code

### Bug report

```cs
                   art  E  JNI ERROR (app bug): accessed stale local reference 0x200001 (index 0 in a table of size 0)
                         F  art/runtime/indirect_reference_table.cc:67] JNI ERROR (app bug): see above.
        google-breakpad  W  ### ### ### ### ### ### ### ### ### ### ### ### ###
                         W  Chrome build fingerprint:
                         W  68.0.3440.91
                         W  344009150
                         W  ### ### ### ### ### ### ### ### ### ### ### ### ###
                  CRASH  E  signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr da8d0000
                         E  *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
                         E  Build fingerprint: 'Verizon/nobleltevzw/nobleltevzw:6.0.1/MMB29K/N920VVRS2BPF2:user/release-keys'
                         E  Revision: '9'
                         E  pid: 32461, tid: 306, name: Thread-2392  >>> com.************.********** <<<
                         E      r0 00000000  r1 00000000  r2 0001f3c4  r3 da8d0000
                         E      r4 da8cdfbc  r5 da8cdf40  r6 e9b443b9  r7 da8ce020
                         E      r8 00000000  r9 da8cdf50  sl c69c9a80  fp da8cdfbc
                         E      ip 00000004  sp da8cded4  lr f612cd61  pc f74b0a98  cpsr eeb61c4c
                         E  backtrace:
                         E      #00  pc 00017a98  /system/lib/libc.so (memset+48)
                         E      #01  pc 00009d5d  /system/lib/libunwind.so (_Uarm_local_access_addr_space_init+8)
                         E      #02  pc e9b4608c  <unknown/absolute>
                         E  stack:
                         E           da8cde94  ffffffff
                         E           da8cde98  ffffffff
                         E           da8cde9c  00000000
                         E           da8cdea0  00000000
                         E           da8cdea4  00000000
                         E           da8cdea8  00000032
[...]

PR #91, opened 2018-08-22
Fix android method call

martin308:
Discovered by @willispinaud in #88, the way we were calling this method fails on android v6. Dropping back down to the lower level JNI calls resolves this. 

Have tested this on android 4, 5 and 6 on a device.

### DIFF
--- a/src/BugsnagUnity/DictionaryExtensions.cs
+++ b/src/BugsnagUnity/DictionaryExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using BugsnagUnity.Payload;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


fractalwrench:
Would it be possible to get an explanation of what the issue was on Android? Is this an issue on later devices e.g. Android 7, 8?

martin308:
It wasn't entirely obvious what the issue was. My biggest hunch is that it couldn't infer which method to call based on the arguments provided anymore. Although I'm not sure how this relates to just the android version changing. I'm currently working on getting unity android maze-runner tests running so that I can test on other versions

fractalwrench:
Adding new tests that cover this functionality, and running them on all API levels sounds like a good start. I think it'd also be worth spending a bit of time trying to understand why the original code was causing a crash, as it may reveal that the fix doesn't work for all inputs, or that similar issues exist elsewhere in our use of the JNI.

PR #92, opened 2018-08-29
Remove allocation in update

dennis-pan-scopely:
## Goal
Remove GC allocations in update function.
bugsnagupatedictionarybefore

### Added
LogTypeComparer to avoid default comparer allocating memory

## Tests
Manually tested in profiler

### DIFF
--- a/src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
+++ b/src/Assets/Standard Assets/Bugsnag/Bugsnag.cs
@@ -325,8 +325,18 @@ public class Bugsnag : MonoBehaviour {
       { LogType.Warning, 5 }
     };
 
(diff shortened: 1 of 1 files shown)


martin308:
@dennis-pan-scopely thanks for the PR! Unfortunately we are in the middle of restructuring this notifier in a major way. You can see the work in progress over [here](https://github.com/bugsnag/bugsnag-unity/pull/83). This has been addressed in the new code, [here](https://github.com/bugsnag/bugsnag-unity/blob/6abdb250b8b2886ede917331f492235fbc89d2e5/src/BugsnagUnity/UniqueLogThrottle.cs#L33). If you see anything in the new code, I'd love to get your thoughts 😎

dennis-pan-scopely:
@martin308 looks great! What's the release schedule for the new version?

PR #93, opened 2018-09-14
Create per platform dlls

martin308:
When using `il2cpp` backend on non cocoa based builds, `il2cpp` does not like the fact that there are calls to native code that does not exist outside of cocoa. To get around this we create a dll for each supported platform and one that is a "fallback" to handle platforms where we don't have any special code to support. By separating all of the platform specific code into the individual dlls `il2cpp` has no problems with the code on non cocoa platforms.

The metafiles for all of the new dlls seems to work on my testing across unity 2017 and 2018 so that each of the platforms that we support are able to be built. The spec for these files is pretty opaque though so it's possible that they could be simpler than they are right now.


martin308:
Merging this down, PR comments still welcome on here and in the main PR

PR #105, opened 2018-10-31
Use same delivery everywhere

kattrali:
## Goal

Reduce surface of JNI code, unifying delivery strategy everywhere

### Removed

* `AndroidDelivery`
* `JavaStreamWrapper`

### Changed

* Use default delivery client in `AndroidClient`

## Tests

* Tested locally against Android 19, 28

## Review

What I need reviewed most is that the change works at the extremes of Android devices:

* API 16 ARM v7a
* ARM 23 ARM v7a
* ARM 23 ARM 64
* ARM 28 x86


fractalwrench:
It'd be good to get some more context on the motivation behind this change, because at the moment I'm not sure why this changeset is necessary.

Why did we originally implement a separate AndroidDelivery mechanism? Why do we need to switch to the C# Delivery mechanism at this stage? Are there any particular risks we should be wary of?

PR #107, opened 2018-11-08
refactor: migrate JNI interaction to use NativeInterface

fractalwrench:
## Goal

Allow capture of background crashes in Android Unity notifier, by registering for `logMessageReceivedThreaded` callbacks.

## Changeset

The majority of interaction with the Android notifier now happens through `NativeInterface.cs`, which
proxies into `NativeInterface.java`. Implementing this has allowed the use of `logMessageReceivedThreaded` rather than `logMessageReceived`, which increases the amount of captured information.

Additionally, `Client.cs` has been altered to check whether a metadata value is already contained in the payload before adding it. This appears necessary due to the following call earlier in its constructor:

`NativeClient.SetMetadata(UnityMetadataKey, UnityMetadata.ForNativeClient());`

## Tests

Ran manual tests on an Android example project with a main thread and background thread crash with the following data, and manually inspected the dashbboard:

- Custom user
- Simple breadcrumb
- Breadcrumb with metadata and different stater
- Nested metadata
- Added metadata

### Linked issues

https://github.com/bugsnag/bugsnag-android/pull/384


kattrali:
Added additional refactor to:

1. Make `Configuration` functions also use the NativeInterface
2. Ensure the JNI is attached to the current thread (and detach if manually attached)

PR #124, opened 2019-01-31
Fix Android ReleaseStage not being set

Cawllec:
## Goal
Fixes an issue where the ReleaseStage wasn't being passed into the Android notifier, causing the incorrect stage to be notified.

## Tests
Should be manually verified on:
- [x] Android
- [x] iOS
- [x] Windows
- [x] OSX

### DIFF
--- a/src/BugsnagUnity/Configuration.cs
+++ b/src/BugsnagUnity/Configuration.cs
@@ -43,8 +43,6 @@ namespace BugsnagUnity
 
     public virtual string AppVersion { get; set; }
 
-    private string RealAppVersion { get; set; }
-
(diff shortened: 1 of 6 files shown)


src/BugsnagUnity/Configuration.cs
```diff
@@ -10,17 +10,19 @@ public abstract class AbstractConfiguration : IConfiguration
 
     public const string DefaultSessionEndpoint = "https://sessions.bugsnag.com";
 
-    public AbstractConfiguration(string apiKey)
+    public AbstractConfiguration() {}
+
+    protected virtual void SetupDefaults(string apiKey)
```
Cawllec: If all the default values that would be overridden are set here, they won't need to be set in the other configurations.

PR #126, opened 2019-02-01
Fix Delivery failure Logging

Cawllec:
## Goal
Ensures that there's a delay between a delivery failing and a retry to avoid log spamming.  Also delays in the case of no network connectivity.

## Test
Attached ZIP features a built package with an extra log included that should trigger in the event of a `500` response, ensuring that the correct delay occurs between delivery attempts.
[Bugsnag.unitypackage.zip](https://github.com/bugsnag/bugsnag-unity/files/2821733/Bugsnag.unitypackage.zip)

### DIFF
--- a/src/BugsnagUnity/IDelivery.cs
+++ b/src/BugsnagUnity/IDelivery.cs
@@ -110,15 +110,20 @@ namespace BugsnagUnity
         {
           // success!
         }
-        // once we can drop support for unity 5.6 we can use req.isNetworkError
-        // instead of req.error != null. According to the unity docs though this
-        // should have the same effect
+        else if (req.responseCode == 404)
+        {
+          DelayBeforeDelivery = true;
+          Send(payload);
+        }
         else if (req.responseCode >= 500)
         {
           // something is wrong with the server/connection, should retry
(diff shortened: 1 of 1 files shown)


src/BugsnagUnity/IDelivery.cs
```diff
@@ -42,7 +44,15 @@ void ProcessQueue()
       {
         try
         {
-          SerializeAndDeliverPayload(Queue.Dequeue());
+          if (Application.internetReachability == NetworkReachability.NotReachable) {
+            Debug.LogWarning("Bugsnag: Network not available, temporarily suspending delivery");
```
kattrali: This log level is unnecessarily high as disconnecting from the network is a common condition. It would be preferable to remove it altogether as it would have the same effect as the current failure log.

src/BugsnagUnity/IDelivery.cs
```diff
@@ -42,7 +44,15 @@ void ProcessQueue()
...
+            System.Threading.Thread.Sleep(10000);
```
kattrali: This duration should be a constant.

PR #127, opened 2019-02-04
feat: Report C/C++ crashes on Android

kattrali:
## Goal

Automatically detect and report C/C++ crashes on Android.

## Changeset

Adds bugsnag-android-ndk as a part of the default installation. No additional configuration is required.

## Tests

Updated the test app to include an NDK crash on the `kattrali/bundle-ndk` branch.

### Linked issues

* Related to bugsnag/docs.bugsnag.com#875
* ~Requires bugsnag/bugsnag-android#427 before release~

[Bugsnag.unitypackage.zip](https://github.com/bugsnag/bugsnag-unity/files/2836898/Bugsnag.unitypackage.zip)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,32 @@
 Upgrading
 =========
 
+## 4.1 to 4.2
+
+4.2.0 adds support for reporting C/C++ crashes in Android code. If you are using
+the Gradle build system to export your app, you will need to add the following
+line to your exported project's build.gradle to enable this functionality:
+
+### Gradle 4.2+
+
+```groovy
+dependencies {
+  // ...
+  // add this line:
+	implementation(name: 'bugsnag-android-ndk-release', ext:'aar')
+}
+```
+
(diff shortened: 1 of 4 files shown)


UPGRADING.md
```diff
@@ -1,6 +1,20 @@
...
+
+4.2.0 adds support for reporting C/C++ crashes in Android code. If you are using
+the Gradle build system to export your app, you will need to add the following
+line to your exported project's build.gradle to enable this functionality:
+
+```groovy
+dependencies {
+  // ...
+  // add this line:
+	compile(name: 'bugsnag-android-ndk-release', ext:'aar')
```
Cawllec: Should this be `implementation(name: 'bugsnag-android-ndk-release', ext:'aar')`, as `compile` is deprecated?

kattrali: Its both; older versions of unity export with an older gradle version by default, I'll add a note to switch based on whether gradle 4.2+ is used.

PR #136, opened 2019-03-07
Add stopSession() and resumeSession() to public API

fractalwrench:
## Goal

Adds the ability to stop and resume sessions. This prevents the handled/unhandled error count from incrementing when a session is in the stopped state, which may be desirable if a user manually tracks sessions and does not care about crashes that occur in the background.

See previous PRs on [Android](https://github.com/bugsnag/bugsnag-android/pull/429) and [Cocoa](https://github.com/bugsnag/bugsnag-cocoa/pull/325) for reference.

## Changeset

- Updated bugsnag-android from v4.11.0 to v4.12.0
- Updated bugsnag-cocoa from v5.17.3 to v5.19.0
- Added `StopSession` and `ResumeSession` to `Bugsnag.cs`
- Added boolean flag to `Session.cs` to track whether a session is stopped or not
- Updated `SessionTracker.cs` so that `CurrentSession` returns null if `StopSession` is called, which prevents session information from being added to the session payload.
- Removed failing `check_package` task from Rakefile

## Tests

- Added unit tests to verify the behaviour of `StartSession`, `ResumeSession`, and `StopSession`.
- Added stub classes to `Tests/BugsnagUnity.Tests` to allow for the construction of a `SessionTracker` in NUnit tests.
- Changed some classes to `internal` and used `[assembly: InternalsVisibleTo("class")]` to make classes visible to unit tests
- Manually verified that session information is not in an error payload when stopping a session, on Android + iOS.

### DIFF
--- a/src/BugsnagUnity/SessionTracker.cs
+++ b/src/BugsnagUnity/SessionTracker.cs
@@ -1,10 +1,17 @@
 ﻿using BugsnagUnity.Payload;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("BugsnagUnity.Tests")]
 
 namespace BugsnagUnity
 {
   public interface ISessionTracker
   {
     void StartSession();
+
+    void StopSession();
+
+    bool ResumeSession();
 
     Session CurrentSession { get; }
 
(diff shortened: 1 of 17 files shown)


src/BugsnagUnity/SessionTracker.cs
```diff
@@ -43,10 +57,38 @@ public void StartSession()
...
+      }
+      return resumed;
+    }
     
     public void AddException(Report report)
     {
-      _currentSession?.AddException(report);
+      var session = _currentSession;
+
+      if (session != null) {
```
kattrali: This null check is the same as `_currentSession?`. [Null operator docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators)

PR #144, opened 2019-04-11
fix: Correctly set notifier name/version for iOS native crashes

kattrali:
Tested by booting an example app and inspecting notifier info in the payload and on the dashboard.

### DIFF
--- a/src/BugsnagUnity.mm
+++ b/src/BugsnagUnity.mm
@@ -7,6 +7,10 @@
 #import "BSG_KSSystemInfo.h"
 #import "BSG_KSMach.h"
 
+@interface Bugsnag ()
++ (id)notifier;
+@end
+
 extern "C" {
   struct bugsnag_user {
     const char *user_id;
@@ -33,7 +37,7 @@ extern "C" {
 
   void bugsnag_setMetadata(const void *configuration, const char *tab, const char *metadata[], int metadataCount);
 
-  void bugsnag_startBugsnagWithConfiguration(const void *configuration);
+  void bugsnag_startBugsnagWithConfiguration(const void *configuration, char *notifierVersion);
 
   void *bugsnag_createBreadcrumbs(const void *configuration);
   void bugsnag_addBreadcrumb(const void *breadcrumbs, char *name, char *type, char *metadata[], int metadataCount);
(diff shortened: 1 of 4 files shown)


src/BugsnagUnity.mm
```diff
@@ -135,8 +139,17 @@ void bugsnag_setMetadata(const void *configuration, const char *tab, const char
...
 
-void bugsnag_startBugsnagWithConfiguration(const void *configuration) {
+void bugsnag_startBugsnagWithConfiguration(const void *configuration, char *notifierVersion) {
   [Bugsnag startBugsnagWithConfiguration: (__bridge BugsnagConfiguration *)configuration];
+  if (notifierVersion != NULL) {
+    NSString *ns_version = [NSString stringWithUTF8String:notifierVersion];
+    id notifier = [Bugsnag notifier];
+    [notifier setValue:@{
+        @"version": ns_version,
+        @"name": @"Unity Bugsnag Notifier",
```
snmaynard: Should be `Bugsnag Unity (Objective-C)`? cc @duncanhewett @bullmo

kattrali: It was `Bugsnag Unity (Cocoa)` before. I made it match what android does now, which makes more sense?

kattrali: If it needs to be `Bugsnag Unity (Cocoa)`, I'll update android to match too.

snmaynard: In the spec it says `Bugsnag <Framework> (Dependency)`. The cocoa notifier seems to be `objective-c` -> https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/BugsnagNotifier.m#L216

snmaynard: The pipeline needs to know about any notifier name changes ofc @mikeewhite too

kattrali: The cocoa notifier gets overwritten as `Bugsnag <Platform> Notifier`: 

https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/BugsnagNotifier.m#L336-L341

kattrali: To clarify this change:

* In the latest released version of bugsnag-unity, it reports the cocoa notifier info for native crashes, which right now is `iOS Bugsnag Notifier` v5.18.0. On Android it uses `Unity Bugsnag Notifier` in all cases.
* In bugsnag-unity v3.x, it reported the name as `Bugsnag Unity (Cocoa)` or `Bugsnag Unity (Android)` and the unity notifier version number.

snmaynard: Lets do the sensible thing™ when comparing the spec, and what the various notifiers and pipeline currently do. Your call @kattrali.

kattrali: I have consulted with The Oracle (@bullmo) and will make a slight adjustment.

PR #146, opened 2019-04-15
Session tracking fixes

kattrali:
Fix subtle bugs in the session tracking implementation.

1. (iOS only) Set the correct date for the start of sessions
2. Ensure that resumed sessions notify the native layer to synchronize the change
3. Skips notifying the native layer of a session change when stopping a session if there is no session


kattrali:
Thank you 🙇‍♀️

PR #157, opened 2019-06-27
fix: Synchronize C# metadata changes with native layer

kattrali:
## Goal

When using `Bugsnag.Metadata.Add()`, any added metadata should be synchronized with the native layer to ensure the values are sent with native crash reports as well as Unity log reports.

Example usage:

```c#
void Start() {
    Bugsnag.Metadata.Add("app", new Dictionary<string, string>(){
      {"buildno", "0.1" },
      {"package", "C" }
    });
}
```

## Changeset

* Added a `INativeClient` ref to the `Metadata` initializer
* Enhanced the `Metadata` class with custom `Add()` and `Remove()` implementations which propagate changes to the native layer
* Switches the `Report.Metadata` class to `Dictionary<string, object>` instead of `Metadata` since it requires no synchronization

Requires bugsnag/bugsnag-android#505

### Artifacts

* [Bugsnag.unitypackage.zip](https://github.com/bugsnag/bugsnag-unity/files/3345158/Bugsnag.unitypackage.zip) (md5: 7ff0149c7d62ab8bf74bac0650aa6190, sha1: 2e234891e3c8217ee677e401b75a481afa9f0aae)
* [Bugsnag-with-android-64bit.unitypackage.zip](https://github.com/bugsnag/bugsnag-unity/files/3345159/Bugsnag-with-android-64bit.unitypackage.zip) (md5: b900643ada1ca9d7d932fd742be45cd0, sha1: 408e9ac22c375d6efc41f664ad1716fdd25a2e26)

## Tests

Test handled and unhandled, native and C# crashes on each:

* [x] Test change on iOS
* [x] Test change on macOS
* [x] Test change on Android

### DIFF
--- a/src/BugsnagUnity/Payload/Metadata.cs
+++ b/src/BugsnagUnity/Payload/Metadata.cs
@@ -1,8 +1,42 @@
 ﻿using System.Collections.Generic;
+using BugsnagUnity;
 
 namespace BugsnagUnity.Payload
 {
   public class Metadata : Dictionary<string, object>, IFilterable
   {
+    private INativeClient NativeClient = null;
+
+    public Metadata() {
+    }
+
+    internal Metadata(INativeClient nativeClient) {
+      NativeClient = nativeClient;
+    }
+
+    public void Add(string section, object newValue) {
+      if (NativeClient != null) {
+        if (newValue is Dictionary<string, string> stringValues) {
+          base.Add(section, stringValues);
+          NativeClient.SetMetadata(section, stringValues);
+        } else if (newValue is Dictionary<string, object> objectValues) {
+          var target = new Dictionary<string, string>();
+          foreach(var pair in objectValues) {
+            target.Add(pair.Key, pair.Value.ToString());
+          }
+          base.Add(section, target);
+          NativeClient.SetMetadata(section, target);
+        }
+      } else {
+        base.Add(section, newValue);
+      }
+    }
+
+    public void Remove(string section) {
+      base.Remove(section);
+      if (NativeClient != null) {
+        NativeClient.SetMetadata(section, null);
+      }
+    }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ### Bug fixes
 
+* Show report metadata added using `Bugsnag.Metadata.Add()` in native crash
+  reports
+  [#157](https://github.com/bugsnag/bugsnag-unity/pull/157)
 * (Android) Fix null pointer dereference when calling Bugsnag.StopSession()
 * Update bugsnag-cocoa dependency to v5.22.2:
   * Fix trimming the stacktraces of handled error/exceptions using the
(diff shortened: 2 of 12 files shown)


src/BugsnagUnity/Payload/Report.cs
```diff
@@ -62,8 +62,8 @@ public void Ignore()
     /// Gets the metadata associated with this report.
     /// </summary>
     /// <value>The metadata.</value>
-    public Metadata Metadata => Event.Metadata;
     
+    public Dictionary<string, object> Metadata => Event.Metadata;
```
fractalwrench: This is a breaking change - I don't fully understand why this needs to change to a dictionary.

src/BugsnagUnity/Payload/Metadata.cs
```diff
@@ -1,8 +1,26 @@
...
   public class Metadata : Dictionary<string, object>, IFilterable
   {
+    private INativeClient NativeClient;
+
+    internal Metadata(INativeClient nativeClient) {
+      NativeClient = nativeClient;
+    }
+
+    public void Add(string section, object newValue) {
+      base.Add(section, newValue);
```
fractalwrench: This will crash if the key already exists in the dictionary, due to how [dictionaries are designed in C#](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.add?view=netframework-4.8#System_Collections_Generic_IDictionary_2_Add__0__1_)

src/BugsnagUnity/Payload/Metadata.cs
```diff
@@ -1,8 +1,26 @@
...
+      if (newValue is Dictionary<string, string> values) {
```
fractalwrench: I'm not sure I follow why the native client's metadata should be updated only if the parameter is a dictionary?

src/BugsnagUnity/Payload/Metadata.cs
kattrali: Correct, and it should since this class is a subclass of Dictionary. I intentionally didn't change this behavior.

src/BugsnagUnity/Payload/Metadata.cs
kattrali: The documented usage of this class is `Metadata.Add(string, dict<string, string>)` which cleanly maps to how metadata is structured. 

If the value is not a dictionary, then there is no representation for storing arbitrary objects and translating them over the native bridge.

src/BugsnagUnity/Payload/Report.cs
kattrali: It isn't - the superclass of `Metadata` is `Dictionary<string, object>` and previously there was no added behavior in the Metadata class. The intent was to separate "Metadata class which syncs with the NativeClient" from "instance of metadata attached to a report", however this could instead switch on whether a native client is set on the instance so I'll do that.

kattrali:
Made a few changes to make the fix less invasive. Metadata now has a non-native-syncing variant.

PR #159, opened 2019-07-01
fix(android): Parse java exception name when message missing

kattrali:
Not all Java exceptions have messages, so the default should be to set the class and java exception parsing style if the Android error class is  detected.

Depends on bugsnag/bugsnag-android#511

### DIFF
--- a/src/BugsnagUnity/Payload/Exception.cs
+++ b/src/BugsnagUnity/Payload/Exception.cs
@@ -133,16 +133,28 @@ namespace BugsnagUnity.Payload
       {
         var errorClass = match.Groups["errorClass"].Value;
         var message = match.Groups["message"].Value.Trim();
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/Payload/Exception.cs
```diff
@@ -141,8 +141,13 @@ public static Exception FromUnityLogMessage(UnityLogMessage logMessage, System.D
           {
             errorClass = match.Groups["errorClass"].Value;
             message = match.Groups["message"].Value.Trim();
-            lines = new StackTrace(logMessage.StackTrace, StackTraceFormat.AndroidJava).ToArray();
           }
+          else
+          {
+            errorClass = message;
```
tomlongridge: Can you add some level of commenting around the format of the AndroidJavaException message?

kattrali:
Updated with some additional comments around how Android Java exceptions are passed through the Unity log handler.

PR #164, opened 2019-09-05
Add overloaded init method with autoNotify param

fractalwrench:
## Goal

Allows disabling of `autoNotify` functionality on Android + iOS.

### DIFF
--- a/src/BugsnagUnity/Native/Android/Configuration.cs
+++ b/src/BugsnagUnity/Native/Android/Configuration.cs
@@ -12,7 +12,7 @@ namespace BugsnagUnity
     {
       var JavaObject = new AndroidJavaObject("com.bugsnag.android.Configuration", apiKey);
       // the bugsnag-unity notifier will handle session tracking
-      JavaObject.Call("setAutoCaptureSessions", autoNotify);
+      JavaObject.Call("setAutoCaptureSessions", false);
       JavaObject.Call("setEnableExceptionHandler", autoNotify);
       JavaObject.Call("setDetectAnrs", false);
       JavaObject.Call("setDetectNdkCrashes", autoNotify);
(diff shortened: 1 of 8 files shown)


src/BugsnagUnity/Native/Fallback/Configuration.cs
```diff
@@ -6,7 +6,7 @@ namespace BugsnagUnity
 {
   class Configuration : AbstractConfiguration
   {
-    internal Configuration(string apiKey) : base() {
+    internal Configuration(string apiKey, bool autoNotify) : base() {
       SetupDefaults(apiKey);
```
kattrali: The value for the `autoNotify` property should be used to set the value for `AutoNotify` in this object for the non-native platforms. It is used to determine whether to capture Unity log messages.

src/BugsnagUnity/Native/Android/Configuration.cs
```diff
@@ -8,13 +8,14 @@ class Configuration : AbstractConfiguration
   {
     internal NativeInterface NativeInterface { get; }
 
-    internal Configuration(string apiKey) : base()
+    internal Configuration(string apiKey, bool autoNotify) : base()
```
kattrali: This implementation of `Configuration` does not override `AutoNotify` so it will also need to be set from the `autoNotify` parameter. The easiest way is probably to pass the value to `setupDefaults` and handle it there.

src/BugsnagUnity/Native/Cocoa/Configuration.cs
```diff
@@ -10,10 +10,11 @@ class Configuration : AbstractConfiguration
   {
     internal IntPtr NativeConfiguration { get; }
 
-    internal Configuration(string apiKey) : base()
+    internal Configuration(string apiKey, bool autoNotify) : base()
```
kattrali: Same as Android, this implementation of `Configuration` does not override `AutoNotify` so it will also need to be set from the `autoNotify` parameter. Otherwise Unity log messages could still generate reports.

fractalwrench:
Thanks for the catch, I've updated to set the C# property as well.

PR #178, opened 2019-11-18
fix(ios): Fix detection of out-of-memory crashes in unity env

kattrali:
By default, `TARGET_OS_*` macros are not defined for static library
targets. Since the OOM feature was enabled based on the flag, the
feature was disabled entirely. The change updates the build
configuration to set the flag conditionally, and link the required
framework(s).

## Testing

* [x] Tested manually by creating native out-of-memeory events before and after the change
* [x] Archived the app successfully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


kattrali:
We have no automated suite for Unity iOS because of some technical hurdles. I added a note that it should be checked during the release process.

PR #184, opened 2020-01-13
Detect ANRs on Android and provide configuration option to enable detection

fractalwrench:
## Changeset

- Adds `AutoDetectAnrs` flag (defaulting to false) to `Configuration`. When this flag is altered, a method is invoked on the native interface which enables/disables ANR detection
- Adds an `AutoDetectAnrs` flag to the `BugsnagBehaviour` script, which shows up as a checkbox in the GUI
- Adds NDK + ANR crash buttons to the example app
- Updates bugsnag-android to allow for the SIGQUIT handler to be disabled:  https://github.com/bugsnag/bugsnag-android/pull/743

### Configuration

`Configuration` is now a publicly accessible interface to allow for users to supply a value for `detectAnrs` up-front. This also brings bugsnag-unity more inline with the notifier specification, although configuration options are still considered mutable after initialisation to remain backwards compatible.

## Tests
- Verified that an ANR is detected (or not) when setting each combination of the `AutoNotify/AutoDetectAnrs` flags in the example app


src/BugsnagUnity/Native/Android/Configuration.cs
```diff
@@ -4,29 +4,32 @@
...
     // Cached value of native-layer auto notify configuration setting to reduce the
     // number of native calls required when reporting a Unity error
     private bool _autoNotify;
+    private bool _autoDetectAnrs = false;
+    private AndroidJavaObject JavaConfig;
 
     internal NativeInterface NativeInterface { get; }
 
-    internal Configuration(string apiKey, bool autoNotify) : base()
+    public Configuration(string apiKey) : base()
```
kattrali: This change introduces an API inconsistency. Calling `Init()` should initialize Bugsnag and I would not expect it to start processing events prior to that point. However, the contents of the `NativeInterface` constructor (used in this method) initialize a Java `Client` object and from that point, Java and C/C++ crashes could be detected and reported. The poor abstraction and separation of concerns here is why the constructor with a boolean to specify autoNotify behavior was added instead of publishing the `Configuration` class. 

The usage of JNI objects also means that this class is not thread-safe, which would be surprising for a seemingly simple C# class and should be at minimum documented, though I suspect this interface change could be better with a more holistic design for initializing with a configuration and having library consumers passing a simpler object around.

fractalwrench: Is there a way out of this without breaking our API compatibility for now? I.e. should the initialisation be altered in this PR to allow passing in a Configuration object without invoking the JNI, or do we need to revert to the old way of doing things?

kattrali: I'm sure there's a few different ways out - maybe sketch out how those look and if one stands out as an obviously better solution, try that? or else we can discuss what the options you come up with. 

A public config object should be safe for a library consumer to manipulate at will and that doesn't need to be internal `Configuration` classes (which could also be renamed). JNI invocation could also be delayed or manually invoked during `Init`.

fractalwrench: I think there are two obvious options:

Option A: use old API that creates `Configuration` object internally during `Bugsnag.init()`. This means a user would have to disable ANR detection after initialisation.

Option B: create an `OnInit()` method on `Configuration` that is invoked at the end of `Bugsnag.init()`, and otherwise make the class publicly accessible like it currently is in this changeset. In `OnInit()` the `NativeInterface` would be constructed, and any `NativeInterface` calls within the `Configuration` class would be gated with a check to see if this method has already run. 

Of these two options I prefer Option A as it's more of a known quantity. We will want to make `Configuration` immutable/not instantiate a native client at some point down the line, but I'm not convinced it's required for this PR. What are your thoughts?

fractalwrench: I have updated the implementation to use the previous API (Option A) as we can address the `Configuration` interface changes at a later date.

PR #187, opened 2020-01-20
Prevent SIGABRT when altering Configuration on background thread

fractalwrench:
## Goal

Prevents a SIGABRT when altering a configuration option such as `Context` or `AppVersion` which accepts a `String` parameter. This behaviour can only be seen on a background thread where a JNIEnv has not been attached in the [`MakeJavaString` method](https://github.com/bugsnag/bugsnag-unity/blob/next/src/BugsnagUnity/Native/Android/NativeInterface.cs#L485).

## Changeset

Updated the `MakeJavaString` implementation to perform a check to see if the JNI is attached, and if not, attach then detach it as in other method implementations such as `CallNativeVoidMethod`.

## Tests

Manually verified that changing the `Context/AppVersion` no longer crashes the app, and that reports can be sent.

**Note:** the `Context/AppVersion` were **not updated** in the report observed on the dashboard. This because existing logic [here](https://github.com/bugsnag/bugsnag-unity/blob/next/src/BugsnagUnity/Native/Android/NativeInterface.cs#L49) prevents a JNIEnv being created for certain CPU architectures.

I have left this behaviour as-is but we should consider documenting this much more clearly, or addressing it as a separate piece of work if this is something we can workaround.


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Prevent SIGABRT when altering Configuration on background thread
```
kattrali: The entry should specify the affected platform

fractalwrench:
Updated implementation to sanitise strings within `CallNativeVoidMethod` instead, as discussed IRL.

PR #190, opened 2020-01-22
Release 4.8.0

fractalwrench:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 4.8.0 (TBD)
+
+* Detect ANRs on Android and provide configuration option to enable detection
```
tomlongridge: We should have an Enhancements subheading here and I would suggest a little more descriptive on how to enable this option. (Looking down the changelog this seems to have been done in the past for new features.)

PR #194, opened 2020-03-24
Avoid using deprecated AndroidJNI API in Unity 2019

fractalwrench:
## Goal

Avoids using a AndroidJniHelper method which was deprecated in Unity 2019. [https://docs.unity3d.com/2019.1/Documentation/ScriptReference/AndroidJNI.FromByteArray.html](AndroidJNI.FromByteArray) was deprecated in favour of a method which takes an [`sbyte`]() instead. This results in a warning being logged to Android logcat whenever the `MakeJavaString` method was invoked, causing lag as the application struggles to write out all the log messages.

## Changeset

Alter the `MakeJavaString` implementation to attempt to use the new API, and fallback to the legacy API if this new API is not available.

Because bugsnag-unity is delivered as a plugin, it is not possible to use the `UNITY_2019_1_OR_NEWER` preprocessor macros to determine whether the new/legacy API should be used. Instead this approach relies on catching an `Exception` thrown when the new constructor cannot be found on earlier versions of Unity, and falling back to use the old constructor instead.

## Tests

Manually verified that no deprecation messages are logged out in Unity 2017.3 and Unity 2019.3, and that both versions are capable of delivering error reports.

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -38,10 +38,13 @@ namespace BugsnagUnity
     private IntPtr ObjectToString;
 
     private bool CanRunOnBackgroundThread;
+
+    private bool Unity2019OrNewer;
     private Thread MainThread;
 
     public NativeInterface(AndroidJavaObject config)
     {
+      Unity2019OrNewer = IsUnity2019OrNewer();
       MainThread = Thread.CurrentThread;
       using (var system = new AndroidJavaClass("java.lang.System"))
       {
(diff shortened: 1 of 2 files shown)


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -38,10 +38,13 @@ class NativeInterface
     private IntPtr ObjectToString;
 
     private bool CanRunOnBackgroundThread;
+
+    private bool Unity2019OrNewer;
```
tomlongridge: Can we word this wrt whether `SBytes` is available rather than the Unity version? I think it makes it a little more obvious what we're doing.

```suggestion
    private bool CanUseSBytes;
```

tomlongridge: Discussed IRL and agreed to leave as-is.

PR #197, opened 2020-06-09
Fix Ndk reference leaks

snmaynard:
The native interface code should really be rewritten to make it easier to see what constructs native references and what is responsible for disposing of them, as well as having a more standardised way of communicating with the JVM from the csharp layer.

This is an attempt to fix the current issues, but a broader rewrite is not far away.

I have also added a change to the example application which shows how I was reproducing the issue.

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -176,7 +176,6 @@ namespace BugsnagUnity
     }
 
     public void SetNotifyReleaseStages(string[] stages) {
-      //TODO:SM This method leaks some references but is not called often
       if (!CanRunJNI()) {
         return;
       }
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -11,16 +11,8 @@ namespace BugsnagUnity
   class NativeInterface
   {
     private IntPtr BugsnagNativeInterface;
-    private IntPtr BugsnagUnityClass;
     // Cache of classes used:
-    private IntPtr BreadcrumbClass;
-    private IntPtr BreadcrumbTypeClass;
-    private IntPtr CollectionClass;
-    private IntPtr IteratorClass;
-    private IntPtr ListClass;
     private IntPtr MapClass;
-    private IntPtr MapEntryClass;
-    private IntPtr SetClass;
     private IntPtr StringClass;
     // Cache of methods used:
     private IntPtr BreadcrumbGetName;
@@ -39,12 +31,10 @@ namespace BugsnagUnity
 
     private bool CanRunOnBackgroundThread;
 
-    private bool Unity2019OrNewer;
     private Thread MainThread;
 
     public NativeInterface(AndroidJavaObject config)
     {
-      Unity2019OrNewer = IsUnity2019OrNewer();
       MainThread = Thread.CurrentThread;
       using (var system = new AndroidJavaClass("java.lang.System"))
       {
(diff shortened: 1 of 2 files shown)


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -54,65 +44,57 @@ public NativeInterface(AndroidJavaObject config)
...
-        jvalue[] args = AndroidJNIHelper.CreateJNIArgArray(new object[] {client});
+        var argsArray = new object[] {client};
+        jvalue[] args = AndroidJNIHelper.CreateJNIArgArray(argsArray);
         AndroidJNI.CallStaticVoidMethod(BugsnagNativeInterface, setClient, args);
+        AndroidJNIHelper.DeleteJNIArgArray(argsArray, args);
 
         // Cache JNI refs which will be used to load report data later in the
         // app lifecycle to avoid repeated lookups
-        var unityRef = AndroidJNI.FindClass("com/bugsnag/android/unity/BugsnagUnity");
-        BugsnagUnityClass = AndroidJNI.NewGlobalRef(unityRef);
```
fractalwrench, line 72: My understanding is that this is required even though it looks like dead code, because it loads the `BugsnagUnity` class which in turn invokes `System.loadLibrary`.

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -454,116 +445,89 @@ private Breadcrumb ConvertToBreadcrumb(IntPtr javaBreadcrumb)
...
-      } catch (System.Exception _) { // use legacy API on older versions
-        return false;
-      }
-    }
-
     private bool CanRunJNI() {
       return CanRunOnBackgroundThread || object.ReferenceEquals(Thread.CurrentThread, MainThread);
     }
 
     private Dictionary<string, object> DictionaryFromJavaMap(IntPtr source) {
```
fractalwrench, line 513: It may be worth investigating whether it's possible to use a combination of [EnsureLocalCapacity](https://docs.unity3d.com/ScriptReference/AndroidJNI.EnsureLocalCapacity.html) and [PushLocalFrame](https://docs.unity3d.com/ScriptReference/AndroidJNI.PushLocalFrame.html) here and in any other methods which require large numbers of local references.

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -355,7 +362,8 @@ private void CallNativeVoidMethod(string methodName, string methodSig, object[]
         var obj = args[i];
 
         if (obj is string) {
-          itemsAsJavaObjects[i] = MakeJavaString(obj as string);
+          //TODO:SM leaking ref
```
fractalwrench, line 365: This leak will need to be resolved

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -379,30 +387,31 @@ private IntPtr CallNativeObjectMethod(string methodName, string methodSig, objec
...
+      AndroidJavaObject[] itemsAsJavaObjects = new AndroidJavaObject[items.Length];
       for (int i = 0; i < items.Length; i++) {
-        itemsAsJavaObjects[i] = MakeJavaString(items[i]);
+        itemsAsJavaObjects[i] = MakeJavaStringRef(items[i]);
       }
       var first = itemsAsJavaObjects[0];
       IntPtr rawArray = AndroidJNI.NewObjectArray(items.Length, StringClass, first.GetRawObject());
       for (int i = 1; i < items.Length; i++) {
         AndroidJNI.SetObjectArrayElement(rawArray, i, itemsAsJavaObjects[i].GetRawObject());
+        itemsAsJavaObjects[i].Dispose();
```
fractalwrench, line 414: This appears to miss a `Dispose()` call on the 0th element

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -454,116 +445,89 @@ private Breadcrumb ConvertToBreadcrumb(IntPtr javaBreadcrumb)
...
-      var name = "<empty>";
-      var nameObj = AndroidJNI.CallObjectMethod(javaBreadcrumb, BreadcrumbGetName, new jvalue[]{});
-      if (nameObj != IntPtr.Zero)
-      {
-        name = AndroidJNI.GetStringUTFChars(nameObj);
-      }
-      var timestamp = AndroidJNI.CallStringMethod(javaBreadcrumb, BreadcrumbGetTimestamp, new jvalue[]{});
+
+      string name = AndroidJNI.CallStringMethod(javaBreadcrumb, BreadcrumbGetName, new jvalue[]{});
+      if(name == null) name = "<empty>";
```
fractalwrench, line 473: Code style is currently not enforced with linters on this repo, but the convention is to use braces for single-line conditionals

PR #198, opened 2020-06-09
Minimal reference leak fixes

snmaynard:
I believe this is the minimal set of changes required to fix most of the reference leaks

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -360,7 +360,7 @@ namespace BugsnagUnity
       }
 
       IntPtr map = CallNativeObjectMethodRef(methodName, "()Ljava/util/Map;", new object[]{});
-      var value = DictionaryFromJavaMap(map);
+      Dictionary<string, object> value = DictionaryFromJavaMap(map);
       AndroidJNI.DeleteLocalRef(map);
 
       if (!isAttached) {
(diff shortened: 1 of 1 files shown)


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -279,9 +287,9 @@ public NativeInterface(AndroidJavaObject config)
       if (!isAttached) {
         AndroidJNI.AttachCurrentThread();
       }
-      using (var map = JavaMapFromDictionary(metadata))
+      using (AndroidJavaObject map = MakeJavaMapDisposable(metadata))
```
fractalwrench, line 290: nit: consider renaming to `MakeDisposableJavaMap` as the current naming implies that `metadata` is of type `java/util/Map`

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -320,26 +331,45 @@ public List<Breadcrumb> GetBreadcrumbs()
...
 
       while (AndroidJNI.CallBooleanMethod(iterator, IteratorHasNext, new jvalue[]{}))
       {
-        var crumb = AndroidJNI.CallObjectMethod(iterator, IteratorNext, new jvalue[]{});
+        IntPtr crumb = AndroidJNI.CallObjectMethod(iterator, IteratorNext, new jvalue[]{});
         breadcrumbs.Add(ConvertToBreadcrumb(crumb));
         AndroidJNI.DeleteLocalRef(crumb);
       }
-
-      AndroidJNI.DeleteLocalRef(javaBreadcrumbs);
```
fractalwrench, line 334: This local ref can still be deleted unless I'm mistaken?

snmaynard: Yeah - I just moved the delete to being before the loop as we dont need a collection reference when using an iterator.

src/BugsnagUnity/Native/Android/NativeInterface.cs
snmaynard: This is the same format as MakeJavaString so its consistent. Maybe Build is better than Make?

src/BugsnagUnity/Native/Android/NativeInterface.cs
fractalwrench: ah, got it - that's fine.

src/BugsnagUnity/Native/Android/NativeInterface.cs
fractalwrench: Build sounds good to me

PR #199, opened 2020-06-09
Add breadcrumb log level from spec

snmaynard:
Adds the breadcrumb log level configuration option from the notifier spec to allow people to control breadcrumb creation.

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -127,7 +127,7 @@ namespace BugsnagUnity
       {
         var logMessage = new UnityLogMessage(condition, stackTrace, logType);
         var shouldSend = Exception.ShouldSend(logMessage)
-          && UniqueCounter.ShouldSend(logMessage) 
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -138,7 +138,7 @@ void Notify(string condition, string stackTrace, LogType logType)
           Notify(new Exception[]{exception}, exception.HandledState, null, logType);
         }
       }
-      else
+      else if (logType.IsGreaterThanOrEqualTo(Configuration.BreadcrumbLogLevel))
```
fractalwrench, line 141: This is inconsistent with the notifier spec. If `LogType.Log` is specified in `config.BreadcrumbLogLevel`, then log-level messages will not be recorded as breadcrumbs - whereas the spec states they should be.

fractalwrench:
The PR description needs updating and a changelog entry needs to be added, otherwise this LGTM.

PR #207, opened 2021-01-22
Respect autoNotify flag on Android

fractalwrench:
## Goal

Fixes Android's handling of the `autoNotify` flag so that it is respected when set as `false` for JVM errors.

## Changeset

`AbstractConfiguration ` set `AutoNotify` to true, which [enabled the JVM exception handler](https://github.com/bugsnag/bugsnag-unity/blob/PLAT-5738/auto-notify-android/src/BugsnagUnity/Native/Android/NativeInterface.cs#L163). This meant that the call to `setEnableExceptionHandler` in the Android implementation of `Configuration` was overridden as soon as the base method of `SetupDefaults()` was called.

This fix passes the autoNotify value supplied through `Bugsnag.Init` to the base method of `SetupDefaults()` so that the handler is not enabled.

## Testing

Manually verified in an example project with `autoNotify = false` that an error is no longer sent.

### DIFF
--- a/src/BugsnagUnity/Configuration.cs
+++ b/src/BugsnagUnity/Configuration.cs
@@ -12,12 +12,12 @@ namespace BugsnagUnity
 
     public AbstractConfiguration() {}
 
-    protected virtual void SetupDefaults(string apiKey)
+    protected virtual void SetupDefaults(string apiKey, bool autoNotify)
     {
       ApiKey = apiKey;
       AppVersion = Application.version;
       AutoCaptureSessions = true;
-      AutoNotify = true;
+      AutoNotify = autoNotify;
     }
 
     public virtual bool ReportUncaughtExceptionsAsHandled { get; set; } = true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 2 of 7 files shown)


fractalwrench:
CI temporarily failing due to issues unrelated to this changeset - Travis CI appears to have changed behind the scenes again

src/BugsnagUnity/Configuration.cs
```diff
@@ -10,14 +10,16 @@ public abstract class AbstractConfiguration : IConfiguration
 
     public const string DefaultSessionEndpoint = "https://sessions.bugsnag.com";
 
+    protected bool _autoNotify = true;
```
tomlongridge, line 13: I'm not keen on how this change leaks out to the abstract class which causes potential confusion with existing private `_autoNotify` variables in the Cocoa implementation.

How about we follow the pattern with the `apiKey` and pass this in as a parameter to `SetupDefaults`?

fractalwrench:
Refactored changeset so that `autoNotify` is passed down to`SetupDefaults`, and confirmed that `AutoNotify` is respected for the original reproduction case.

It may be worth consider some general regression testing in this area, as the changeset now touches all the platforms after the refactor.

tomlongridge (approved):
LGTM - as discussed, manual regression test on Cocoa to follow to check autoNotify is unaffected.

PR #208, opened 2021-01-25
Fix homebrew script

fractalwrench:
## Goal

Fixes the installation scripts for CI, which has started failing due to a breaking change in homebrew. See https://github.com/Homebrew/discussions/discussions/340

This also sets the OSX image to XCode 11.3 for now as this uses Python 2, whereas XCode 11.6 was failing to link brew packages due to using Python 3.


twometresteve (approved):
LGTM

PR #222, opened 2021-03-26
Fix leaks from manual JNI string conversions

kstenerud:
## Goal

The invalid UTF-8 C# to Java conversion crash fix involved manually creating Java native strings, which were leaking. This PR ensures they get cleaned up.

## Testing

Ran repro case manually (create tons of breadcrumbs on multiple threads) before and after changes to verify the fix.

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -398,6 +398,36 @@ namespace BugsnagUnity
       return value;
     }
 
+    // Manually converts any C# strings in the arguments, replacing invalid chars with the replacement char..
+    // If we don't do this, C# will coerce them using NewStringUTF, which crashes on invalid UTF-8.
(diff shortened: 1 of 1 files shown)


fractalwrench (changes requested):
This will reintroduce a unicode encoding crash that was fixed in #165 (also https://github.com/bugsnag/bugsnag-android/pull/584). The solution to this problem needs to call `BuildJavaStringDisposable`, invoke the JNI method, and then dispose of the object.

kstenerud:
In that case, shouldn't we also fix the other CallNativeXYZ methods? They just pass the args directly as well...

fractalwrench:
It'd be worth checking the methods again but in practice I don't think they're affected as they don't pass in user-submitted args.

fractalwrench (approved):
LGTM assuming this has been retested and the crash no longer occurs

kstenerud:
Yup, tested with Abigail's repro case project.

PR #225, opened 2021-04-20
Add bugsnag prefix to namespace of vendored SimpleJson

fractalwrench:
## Goal

Adds a bugsnag prefix to the vendored `SimpleJson.cs`, to avoid symbol clashes if a user has vendored `SimpleJson` themselves.

## Testing

Verified on [the following branch](https://github.com/bugsnag/bugsnag-unity/compare/simple-json-repro) that an example app with a vendored `SimpleJson` class failed to build without these changes, and builds with the changes.

### DIFF
--- a/src/BugsnagUnity/SimpleJson.cs
+++ b/src/BugsnagUnity/SimpleJson.cs
@@ -67,12 +67,12 @@ using System.Globalization;
 using System.Reflection;
 using System.Runtime.Serialization;
 using System.Text;
-using SimpleJson.Reflection;
+using BugsnagUnity.Reflection;
 
 // ReSharper disable LoopCanBeConvertedToQuery
 // ReSharper disable RedundantExplicitArrayCreation
 // ReSharper disable SuggestUseVarKeywordEvident
-namespace SimpleJson
+namespace BugsnagUnity
 {
     /// <summary>
     /// Represents the json array.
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/SimpleJson.cs
```diff
@@ -1636,7 +1636,7 @@ private static bool CanAdd(MemberInfo info, out string jsonKey)
 
 #endif
 
-    namespace Reflection
```
kattrali, line 1639: Does this actually need to be changed, since its in a nested namespace?

src/BugsnagUnity/SimpleJson.cs
```diff
@@ -465,7 +465,7 @@ public override IEnumerable<string> GetDynamicMemberNames()
     }
 }
 
-namespace SimpleJson
```
kattrali, line 468: What about `Bugsnag`or `BugsnagUnity` here, to be consistent with the rest of the package? It would also reduce any developer confusion in intellisense to start typing `Bugsnag…` and get `…SimpleJson`.

src/BugsnagUnity/SimpleJson.cs
fractalwrench: This was not required - I've kept the namespace as is.

src/BugsnagUnity/SimpleJson.cs
fractalwrench: Updated to use `BugsnagUnity` for consistency with the rest of the project, then retested against the repro case.

kattrali (approved):

PR #228, opened 2021-04-22
Create universal implementation of Configuration class

fractalwrench:
## Goal

Creates a universal implemetation of `Configuration` by removing the platform-specific implementations. This achieves greater consistency between platforms and allows for the object to be mapped onto the native `Configuration` object when creating a native client.

## Changeset

- Removed platform-specific implementations, and renamed `AbstractConfiguration` to `Configuration`
- Removed `Configuration` 
- `NativeClient` is now responsible for starting the Android/Cocoa client
- The C# configuration object is used to construct an Android/Cocoa `Configuration` object during initialization
- Setting native configuration options after Bugsnag is started has no effect on notifier behaviour
- Removed accessor for `Bugsnag.Configuration` to enforce supplying `Configuration` up-front

## Testing

Ran the example app on the following platforms to confirm that errors are sent:

- Android
- WebGL
- iOS
- macOS

Additionally improved a unit test to check the default configuration values.

Restructured E2E test fixture and updated scenarios to work with the new API.


kstenerud (approved):

twometresteve (approved):
Test changes LGTM.

PR #231, opened 2021-04-26
Update android build script

fractalwrench:
## Goal

Updates the bugsnag-android-unity to use the latest versions of Gradle/AGP and to pin a specific version of the NDK.

## Testing

Relied on existing E2E test coverage to confirm the package builds correctly.


twometresteve (approved):
LGTM

PR #234, opened 2021-04-30
Update to use cocoa v6

kstenerud:
## Goal

PLAT-6369 and 6370

Update the unity plugin to use the latest cocoa library (major version 5 -> 6)

### Changeset

* Update build script to build from the latest cocoa library
* Modified native bridge calls to account for cocoa API changes

## Testing

Manual verification for now until e2e tests are available.

### DIFF
--- a/src/BugsnagUnity.mm
+++ b/src/BugsnagUnity.mm
@@ -4,13 +4,13 @@
 #import "BugsnagUser.h"
 #import "BugsnagNotifier.h"
 #import "BugsnagSessionTracker.h"
+#import "Bugsnag+Private.h"
+#import "BugsnagClient+Private.h"
 #import "BSG_KSSystemInfo.h"
 #import "BSG_KSMach.h"
 #import "BSG_KSCrash.h"
-
-@interface Bugsnag ()
-+ (id)notifier;
-@end
+#import "BSG_RFC3339DateTool.h"
+#import "BugsnagBreadcrumb+Private.h"
 
 extern "C" {
   struct bugsnag_user {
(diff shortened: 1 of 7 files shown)


src/BugsnagUnity.mm
```diff
@@ -77,25 +74,30 @@ void bugsnag_setReleaseStage(const void *configuration, char *releaseStage) {
...
+    if (releaseStage != nil) {
+      NSString *ns_releaseStage = [NSString stringWithUTF8String: releaseStage];
+      [ns_releaseStages addObject: ns_releaseStage];
+    }
   }
-  ((__bridge BugsnagConfiguration *)configuration).notifyReleaseStages = ns_releaseStages;
+  ((__bridge BugsnagConfiguration *)configuration).enabledReleaseStages = ns_releaseStages;
 }
 
 void bugsnag_getNotifyReleaseStages(const void *configuration, const void *managedConfiguration, void (*callback)(const void *instance, const char *releaseStages[], int size)) {
```
fractalwrench, line 88: It looks like this method can be deleted as it's now unused, along with a few other similar ones. Here's the list of what I could spot:

- bugsnag_getApiKey
- bugsnag_getReleaseStage
- bugsnag_getNotifyReleaseStages
- bugsnag_getAppVersion
- bugsnag_getContext
- bugsnag_getNotifyUrl

fractalwrench (commented):
I wasn't able to get the example app running with this changeset. If that and the CI failure can be resolved then the changes look like they're pretty much there - I left one minor comment around unused functions but that's it.

fractalwrench:
Added two commits to fix the test assertions - I'm happy with the code changes as they are.

fractalwrench (approved):
LGTM - discussed test changes with Karl and he's happy with my alterations.

PR #235, opened 2021-04-30
Mobile targets test fixture foundation

twometresteve:
## Goal

Adds the basis of a test fixture for mobile platforms.

## Design

At this stage the test ficture must be manually built and local IP address manually set - but it provides the foundation for some semi-automated tests to be written.

## Changeset

Presenting this as a PR at this stage as most of it is boilerplate and not doesn't require a lot of attention now it's working.  Subsequent PRs will b more interesting.

## Testing

Manually tests locally with a simple HTTP server that prints our received error reports from the test fixture.

PR #236, opened 2021-05-03
Proof of concept test automation for mobile

twometresteve:
## Goal

Follows on from #235 .  Proof of concept scenario for mobile testing.

## Design

Uses a very rough and ready approach to pushing buttons on the screen at fixed (x, y) coordinates.  The positions are wholly dependent on the size of device screen and therefore not intended to be run against a breadth of devices at this stage.

## Changeset

New PoC test scenario and associated Maze Runner setup.

## Testing

Scenario passes when run locally after build the test fixture manually - details to follow in a refreshed `TESTING.md`.


test/mobile/features/steps/unity_steps.rb
```diff
@@ -0,0 +1,32 @@
+When("I wait for the game to start") do
+  # Wait for a fixed time period
+  sleep 10
+end
+
+When("I tap the {string} button") do |button|
+  # TODO Currently specific to the ANDROID_9_0 device (Google Pixel 3)
```
fractalwrench, line 7: I did a little bit of research and think the Unity approach for supporting other Android devices by using [Canvas with screen space overlay enabled](https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/UICanvas.html#screen-space---overlay) to draw the buttons (which I think we already do), and [Screen](https://docs.unity3d.com/ScriptReference/Screen-height.html) to get the width/height/dpi. 

The Appium approach (which might be better) looks like it would involve getting the [window size](https://appium.io/docs/en/commands/web/window/get-window-size/) - and then we could try and alter the UI in Unity so that it's always a fixed percentage of the screen.

fractalwrench (approved):
This LGTM but I think it'd be helpful to add docs to `TESTING.md` before merging down - I wasn't entirely sure how I'd go about running the tests as-is.

test/mobile/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem "bugsnag-maze-runner", git: "https://github.com/bugsnag/maze-runner", tag: "v5.0.1"
```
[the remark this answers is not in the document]
twometresteve: Yes, we are hoping to do that next - there are a few confusing things about the existing macOS tests for someone used to later Maze Runner setups.

test/mobile/features/steps/unity_steps.rb
twometresteve: There's certainly a lot of scope for looking at how we take this forward and it's likely to be a team effort on the approach.

twometresteve:
> This LGTM but I think it'd be helpful to add docs to `TESTING.md` before merging down - I wasn't entirely sure how I'd go about running the tests as-is.

I was going to add a TESTING.md here, but it's now overtaken by events and the TESTING.md will appear in my very next PR - where I've added building of the test fixture and running of the tests to the CI pipeline.

PR #238, opened 2021-05-05
Placeholder e2e tests for Android target

twometresteve:
## Goal

Automate the building of the Android end-to-end test fixture and running the tests.

## Design

The pipeline takes an approach in line with our other notifiers.  Test fixtures are built first in Buidlkite step and attached as build artifacts for consuming in the steps that run the e2e tests using Maze Runner. 

## Changeset

New pipeline steps to build test fixture for Unity 2017 and 2018, backed by a shell script and C# file class to perform the build.

## Testing

Covered by CI.


.buildkite/pipeline.yml
```diff
@@ -91,3 +91,122 @@ steps:
...
+
+  - label: ':android: Build Android test fixture for Unity 2019'
+    skip: Pending PLAT-6447
+    key: 'build-android-fixture-2019'
+    depends_on: 'build-artifacts'
+    timeout_in_minutes: 30
+    agents:
+      queue: opensource-mac-unity
+    env:
+      UNITY_VERSION: "2019.4.25f1"
```
fractalwrench, line 181: This doesn't need action now and is more of a general point, but it'd be nice across the pipelines if there was an easier way to generate this sort of YAML from a centralized test matrix of different versions. For example, on the Android pipeline there are lots of jobs running E2E tests which are effectively identical but target different devices.

test/mobile/features/scripts/build_fixture.sh
```diff
@@ -0,0 +1,47 @@
...
+# package manager (we just don't need it and it slows things down)
+DEFAULT_CLI_ARGS="-quit -batchmode -logFile unity.log -noUpm"
+project_path=`pwd`/maze_runner
+
+# Creating a new project in the MyProject directory
+Unity $DEFAULT_CLI_ARGS -createProject $project_path
+
+# Installing the Bugsnag package
+echo "Importing Bugsnag.unitypackage into $project_path"
+Unity $DEFAULT_CLI_ARGS -projectPath $project_path -importPackage $script_path/../../../../Bugsnag.unitypackage
```
fractalwrench, line 28: `Unity` isn't in my path - is there any way this could work like the way Rakefile calculates the location of Unity Hub?

fractalwrench (approved):
LGTM - although I wasn't able to build the test fixture out of the box so it'd be good to resolve that either through a change to the script or docs.

test/mobile/features/scripts/build_fixture.sh
twometresteve: I'll follow this PR up straight away to make the build Rake driven, so it should make better use of the existing intelligence.

.buildkite/pipeline.yml
twometresteve: Agree and I've raised epic PLAT-6553 to look at dynamically generated pipelines.

PR #239, opened 2021-05-05
PLAT-5159 Unity TimingTrackerObject destroys on load

rich-bugsnag:
## Goal

The session tracking object was not persisting across scenes

## Design

Adding DontDestroyOnLoad is the cleanest way to make sure it stays around. 
I would have added a singleton pattern to make sure it was the only one created, but i checked and the init process that creates it already has a check to make sure init only runs once.

## Changeset

Added DontDestroyOnLoad to the awake method of the TimingTrackerObject behaviour

## Testing

This was tested manually in the unity editor


kattrali (approved):
nice one

PR #240, opened 2021-05-05
Ensure plugin export leave working tree clean

twometresteve:
## Goal

Remove files and update `.gitignore` to avoid working tree appearing dirty after running `rake plugin:export`.

## Testing

Covered by CI.


richardelms (approved):
Looks ok to me, as those files are generated in the build there is no reason to not ignore them.

rich-bugsnag (approved):

PR #241, opened 2021-05-06
Update mazerunner instructions

fractalwrench:
## Goal

Updates the instructions for running mazerunner so that it's possible to run from just using the contributing guide. The old task has also been removed from the Rakefile to avoid any potential source of confusion.


twometresteve (approved):
lgtm

PR #242, opened 2021-05-07
Create static project for desktop test fixture

twometresteve:
## Goal

Provide a static project for the desktop test fixture instead of dynamically generating it, reducing the number of moving parts for engineers.

## Design

Providing a static project allows us to present a more conventional project layout and is more intuitive to those less familiar with Unity.

## Changeset

The changes in `build_maze_runner.sh` (renamed from `create_unity_project.sh` are a good entry point for reviewing this change set.  The last step calls into the new class in `Builder.cs`, rather than using the command line.  This will be consistent with the approach for building on mobile platforms once added, where is is not possible to build purely from the command line.

The majority of the changes after that are the addition of Unity `.asset` files, which are generated by the platform and do not require specific review of their contents.

## Testing

Covered by CI and I've verified that the plugins and desktop test fixture can be built locally, as well as running the desktop MazeRunner tests.


rich-bugsnag (approved):
looks good to me

test/desktop/features/scripts/build_maze_runner.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+export PATH="/Applications/Unity/Hub/Editor/$UNITY_VERSION/Unity.app/Contents/MacOS:$PATH"
+
+# Set project_path to the repo root
+pushd "${0%/*}"
+  pushd ../../../..
+    package_path=`pwd`
+    echo "Expecting to find Busnag package in: $package_path"
```
Cawllec, line 9: Typo
```suggestion
    echo "Expecting to find Bugsnag package in: $package_path"
```

test/desktop/features/scripts/build_maze_runner.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "Importing $package_path/Bugsnag.unitypackage into $project_path"
+    Unity -nographics -quit -batchmode \
+      -logFile $log_file \
+      -projectPath $project_path \
+      -ignoreCompilerErrors \
+      -importPackage "$package_path/Bugsnag.unitypackage"
+    RESULT=$?
+    if [ $RESULT -ne 0 ]; then exit $RESULT; fi
+
+    cp -R NativeCrashy.bundle maze_runner/Assets/Plugins/OSX/NativeCrashy.bundle
```
Cawllec, line 24: Is there a reason this file can't already exist at that location?

Cawllec (approved):
Added a couple of comments, but nothing that should stop a merge going ahead.

test/desktop/features/scripts/build_maze_runner.sh
twometresteve: Fixed

test/desktop/features/scripts/build_maze_runner.sh
twometresteve: You're right - I've `git mv` moved into place and removed the line from the script.  It means building the test fixture manually with the editor is a more intuitive, 2-step, process - import the notifier and build.

PR #246, opened 2021-05-11
feat (NativeInterface): add SetAutoDetectAnrs method to Bugsnag interface

rich-bugsnag:
## Goal

To provide a way to enable and disable anr reporting on the fly without using the config.

## Design

Just a simple addition to the interface. I did a check with the team if it was ok to have dummy implementations in the platforms that don't have anrs and was given the go ahead.

## Changeset

I extended the interface and made the method publicly available in the Bugsnag class

## Testing

I tested manually with unity 2018 and android simulator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Add SetAutoDetectAnrs method to Bugsnag interface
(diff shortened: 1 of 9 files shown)


fractalwrench (approved):
LGTM

PR #249, opened 2021-05-12
Update committed files to avoid dirtying working directory

twometresteve:
## Goal

Update the `.meta` files committed to git to avoid dirtying the tree when running `rake plugin:export`.

## Design

It looks like this crept in after the native Cocoa notifier was updated.

## Testing

Tested locally to verify that the tree is clean after building the notifier.


kstenerud:
Could we instead add them to the gitignore since they get regenerated anyway? Otherwise we have to keep updating them...

twometresteve:
> Could we instead add them to the gitignore since they get regenerated anyway? Otherwise we have to keep updating them...

I don't think I mind and just opted to follow the existing pattern.  @rich-bugsnag , is it likely to matter either way?

PR #252, opened 2021-05-12
Convert mobile test fixture to static unity project

twometresteve:
## Goal

Set up a static project for the mobile test fixture rather than dynamically creating one for each build.

## Design

This brings the setup for mobile into line with that of desktop and makes it easier to work with the project manually in the Unity editor.

## Changeset

There are lots of changes showing in the diff for this PR, but the vast majority are file moves and files generated by Unity that do not require explicit review.  The interesting changes at those to the `.cs` and `.sh` files.

## Testing

Covered by CI.


.buildkite/pipeline.yml
```diff
@@ -107,6 +107,7 @@ steps:
           - Bugsnag-with-android-64bit.unitypackage
         upload:
           - test/mobile/features/fixtures/maze_runner/mazerunner_2017.4.40f1.apk
+          - test/mobile/features/fixtures/unity.log
```
fractalwrench, line 110: good idea 👍

fractalwrench (approved):
LGTM

PR #254, opened 2021-05-14
Build test fixtures for iOS in CI pipeline

twometresteve:
## Goal

Build the iOS IPAs for e2e testing as part of CI.

## Design

Building an iOS app with Unity is a two step process - generate an Xcode project and then build that.  These steps have been implemented as separate Buidlkite steps to reduce the amount of time spent on build servers with a Unity license (which we have a limited number of).

## Changeset

Note that lot of new files have been generated by Unity that do not require detailed review.

New Rake target for iOS and existing script refactored to separate iOS and Android.  New Pipeline steps to perform the builds.

## Testing

Covered by CI and tested that the new rake target works locally on macOS 10.15 with Xcode 11.  The target fails with Xcode 12, simply because the IPA is generated with a different name.  It should be possible to ensure that the same name results in both versions - to follow in a subsequent PR.


fractalwrench (commented):
Confirmed that the Android test fixture still works. Building the iOS app locally fails for me with the following:

>Error building player because build target was unsupported

That might be down to my local version of XCode/using Big Sur.

twometresteve:
> Confirmed that the Android test fixture still works. Building the iOS app locally fails for me with the following:
> 
> > Error building player because build target was unsupported
> 
> That might be down to my local version of XCode/using Big Sur.

I think we'll need to raise a separate ticket to investigate that.  So far it's building ok for me on 10.15 and one of our 2 CI boxes (looking to fix the 2nd now).

fractalwrench (approved):
LGTM - approving on the basis that it works on CI, and Big Sur support can be skipped for now.

PR #255, opened 2021-05-14
fix(Config.context): Setting context on Unity does not update C# errors

rich-bugsnag:
## Goal

Scene changes were overriding the manually set context values

## Design

It's a simple flag setup that fixes the current setup

## Changeset

Added a bool that is set ot true when Bugsnag.SetContext is called. Then check for that bool in the Scene loaded listener

## Testing

Tested manually in editor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 * Add SetAutoDetectAnrs method to Bugsnag interface
   [#246](https://github.com/bugsnag/bugsnag-unity/pull/246)
+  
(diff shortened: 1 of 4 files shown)


fractalwrench (commented):
There's one edge case where I think the context could still get overwritten, otherwise this looks ready to go:

```
var cfg = new Configuration("api-key");
cfg.context = "custom text";
Bugsnag.Start(cfg);
```

fractalwrench (approved):
LGTM

PR #257, opened 2021-05-17
Add extra Android E2E test scenarios

fractalwrench:
## Goal

Adds some additional E2E test coverage for Android. 

## Changeset

- Added scenario for an uncaught JVM exception
- Added scenario for a raised NDK signal
- Added a step to relaunch the Unity app as a raised signal terminates the process

## Test fixture reorganization

- Altered test fixture so that it assembles native Android code each time by copying an AAR into the plugin directory (this was not working as-is as the test fixture just threw `ClassNotFoundException`)
- Split test fixture code so that the iOS plugin is built from including an xcodeproj in the `Assets` directory (no change)
- Removed .meta files from VCS
- Bumped Android gradle build to use latest versions


Cawllec (approved):
LGTM, pending passing tests

PR #260, opened 2021-05-18
Update example app to include more android/cocoa examples

fractalwrench:
## Goal

Updates the example app so that it includes more examples of errors that can be triggered in Android/Cocoa. This should aid testing and provide better examples for users.

## Changeset

- Bumped android build tools/gradle dependency to modern versions
- Added native signal/C++ exception examples
- Added a wider variety of `Debug.Log` examples
- Reworded button text, symbol names, and exception messages for more clarity on what each action does
- Logged a message if an error is not supported on the target platform (e.g. when clicking the ANR button on WebGL)
- Made the `Main.unity` scene open by default

Screenshot 2021-05-18 at 15 35 35

## Remaining work

The native code for the Cocoa OOM, App Hang, and C++ exception examples has not been implemented yet due to time constraints - this is indicated by TODO comments.

The example app needs to be tested out on the iOS/macOS/WebGL platforms to confirm that the errors are received as expected.

## Testing

I've confirmed that all errors are sent via the Android app.


rich-bugsnag (approved):
LGTM

PR #268, opened 2021-05-21
test: support different .ipa outputs in MR build

tomlongridge:
## Goal

Xcode 11 and Xcode 12 output different .ipa names - the product name "maze_runner" in Xcode 11 and the bundle name "Unity-iPhone" in Xcode 12. We move it to a versioned path anyway, so a simple fix is just to wildcard the `mv` command.

## Testing

Run this in a dummy set of files as I don't have e2e tests running locally.

### DIFF
--- a/test/mobile/features/scripts/build_ios.sh
+++ b/test/mobile/features/scripts/build_ios.sh
@@ -6,6 +6,9 @@ popd
 pushd "$script_path/../fixtures"
 project_path=`pwd`/maze_runner
 
+# Clean any previous builds
+find $project_path/output/ -name "*.ipa" -exec rm '{}' \;
+
 # Archive and export the project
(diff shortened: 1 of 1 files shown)


twometresteve:
This seems like a reasonable approach, but I think it make us more vulnerable to picking up IPAs from previous builds.  I'd be inclined to add something like the following before the build starts (not tested myself!):
```
find $project_path/output/ -name "*.ipa" -exec rm '{}' \;
```

I'd also want to check what happens in the build fails and no IPA is produced, although I guess that case would be caught pretty quickly, even if not by this script directly.

tomlongridge:
Added the clean step but, as you say, I didn't see much improvement in catching the case with no .ipa as it will fail soon after.

twometresteve (approved):
LGTM pending CI

PR #269, opened 2021-05-21
Update tests to build against Unity 2019

Cawllec:
## Goal

Fixes the issue that meant the Unity 2019 test fixture was building with the `-noUpm` flag, which prevented test fixture elements from being correctly recognised (presumably due to missing dependencies).

Removes this flag which may lead to a potential slow-down of builds.  If that is the case a more intelligent approach to applying the flag will be used.


test/mobile/features/scripts/build_android.sh
```diff
@@ -14,9 +14,13 @@ popd
...
 pushd "$script_path/../fixtures"
 
-# Run unity and immediately exit afterwards, log all output, disable the
-# package manager (we just don't need it and it slows things down)
-DEFAULT_CLI_ARGS="-quit -batchmode -logFile unity.log -noUpm"
+# Run unity and immediately exit afterwards, log all output
+DEFAULT_CLI_ARGS="-quit -batchmode -logFile unity.log"
+# Disable the package manager only on Unity 2018 due to dependency clashes
+if [[ "$UNITY_PATH" == *"2018"* ]]; then
+  DEFAULT_CLI_ARGS += " -noUpm"
```
twometresteve, line 21: The spaces around `+=` are causing the build to fail (but TIL that you can use `+=` in bash).

Cawllec: Apparently you can't despite my sources, although I've update this now

test/mobile/features/fixtures/maze_runner/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.4.40f1
+m_EditorVersion: 2018.4.34f1
```
twometresteve, line 1: I didn't know if this could have a negative impact on the 2017 build, but it's built ok and the tests passed, so I guess all is ok.

.buildkite/pipeline.yml
```diff
@@ -323,7 +322,6 @@ steps:
       - rake test:ios:build_xcode
 
   - label: ':ios: Generate test fixture Xcode project for Unity 2019'
-    skip: Pending PLAT-6447
```
twometresteve, line 326: I've just noticed that the next step (that actually builds the fixture) is still skipped.  Given that we don't actually run the tests yet that's perhaps less of an issue - it depends on out appetite for incurring another CI run at this stage.

twometresteve (approved):
LGTM

PR #273, opened 2021-05-24
Fix example app

fractalwrench:
## Goal

The example app couldn't be loaded in earlier versions of Unity. This reverts unnecessary style changes made in #260 and keeps only the necessary functional alterations.

## Testing

Confirmed that errors are sent when building the example app in Unity 2018 for Android.


Cawllec (approved):
LGTM, have confirmed the example app builds and runs on iOS

PR #274, opened 2021-05-25
Bugsnag behaviour refactor

rich-bugsnag:
## Goal

The previous design was broken and over complicated

## Design

This is the fastest and cleanest solution for now

## Changeset

Removed unnecessary editor serialisation code 

## Testing

Manually tested on android and ios

**OLD:**

image

**NEW:**

image


kattrali (approved):
Looks good! Much simpler :tada:

src/Assets/Standard Assets/Bugsnag/BugsnagBehaviour.cs
```diff
@@ -10,144 +10,108 @@
...
+        public bool AutoDetectAnrs = false;
+
+        [Tooltip("Should Bugsnag automatically collect data about sessions.")]
+        public bool AutoCaptureSessions = true;
+
+        public LogType NotifyLevel = LogType.Exception;
+
+        [Header("Advanced Settings")]
+
+        [Tooltip("The number seconds required between unique Unity logs that Bugsnag will convert to breadcrumbs or report as errors(if configured). Increase the value to reduce the number of logs sent.")]
```
fractalwrench, line 43:
```suggestion
        [Tooltip("The number seconds required between unique Unity logs that Bugsnag will convert to breadcrumbs or report as errors (if configured). Increase the value to reduce the number of logs sent.")]
```

fractalwrench (approved):

rich-bugsnag:
image
@tomlongridge @kattrali When trying to serialise the name of this field like it used to be then it adds a space. Can we have it as AutoDetectAnrs instead?

kattrali:
Can you use an attribute or something to fix the space?

rich-bugsnag:
image
created a custom attribute for the ANRs label

PR #275, opened 2021-05-25
fix(maxbreadcrumbs): set the max breadcrumbs in the native layer on init

rich-bugsnag:
Fix max breadcrumbs config setting not working

The native layers where not informed of the config setting

Simply following the setup of other config values 

Added missing code to create native config methods 

Tested manually


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -193,10 +193,12 @@ public NativeInterface(IConfiguration cfg)
...
       }
 
-      // set version/context
+      // set version/context/maxbreadcrumbs
       obj.Call("setAppVersion", config.AppVersion);
       obj.Call("setContext", config.Context);
-      return obj;
+      obj.Call("setMaxBreadcrumbs", config.MaximumBreadcrumbs);
+
+    return obj;
```
fractalwrench, line 201:
```suggestion
      return obj;
```

src/BugsnagUnity/Native/Cocoa/NativeClient.cs
```diff
@@ -33,7 +33,7 @@ public NativeClient(IConfiguration configuration)
       NativeCode.bugsnag_setReleaseStage(obj, config.ReleaseStage);
       NativeCode.bugsnag_setAppVersion(obj, config.AppVersion);
       NativeCode.bugsnag_setNotifyUrl(obj, config.Endpoint.ToString());
-
+      NativeCode.bugsnag_setMaxBreadcrumbs(obj, config.MaximumBreadcrumbs);
```
fractalwrench, line 36:
```suggestion
      NativeCode.bugsnag_setMaxBreadcrumbs(obj, config.MaximumBreadcrumbs);

```

fractalwrench (approved):
LGTM on the assumption that this has been tested on both Android/iOS example apps.

fractalwrench:
Agreed to hold off merging this until v5 is released

PR #277, opened 2021-05-25
chore(test setup) update cocoa sub module and remove linker setting in build script

rich-bugsnag:
## Goal

Needed update to fix manual linker flag issue in test fixtures

## Changeset

Update cocoa sub mod to point at next, removed line in build script for adding the linker flag

## Testing

Manually built and tested on iphone and macos


fractalwrench (approved):
LGTM pending CI

PR #280, opened 2021-05-26
PLAT-6631 Enable test fixture and testing building on Unity 2020

Cawllec:
## Goal

- Uses a different manifest to install Unity 2019/2020 packages depending on the UNITY_VERSION defined.
- Updates the pipeline to build and run test fixtures for the Android, MacOS, and iOS platforms across all supported Unity versions


twometresteve (commented):
Looks like a good approach to me - just the failing macOS tests for 2020 to consider.  Looking at just a couple of scenarios, it looks like we may want to skip the tests for now and raise a ticket for one the notifier engineers to look at?

Cawllec:
After some discussion we believe the issues have been resolved and guarded against happening in the future.

test/mobile/features/scripts/generate_xcode_project.sh
```diff
@@ -16,7 +16,21 @@ pushd "$script_path/../fixtures"
 
 # Run unity and immediately exit afterwards, log all output, disable the
 # package manager (we just don't need it and it slows things down)
```
fractalwrench, line 18: This comment appears to be outdated

test/mobile/features/scripts/generate_xcode_project.sh
```diff
@@ -16,7 +16,21 @@ pushd "$script_path/../fixtures"
...
+  mkdir -p maze_runner/Packages
+  cp manifests/manifest_2020.json maze_runner/Packages/manifest.json
+elif [[ "$UNITY_PATH" == *"2019"* ]]; then
+  # Copy the 2019 package manifest in for that version
+  mkdir -p maze_runner/Packages
+  cp manifests/manifest_2019.json maze_runner/Packages/manifest.json
+else
+  # On all other versions disable the package manager
+  DEFAULT_CLI_ARGS="${DEFAULT_CLI_ARGS} -noUpm"
+fi
```
fractalwrench, line 32: Is there any way to avoid adding duplication between this script and `build_android.sh`/`build_maze_runner.sh`?

Cawllec: Yes, but not without a small re-write of them both.  Regarding both of your queries a couple of tickets, one for the multi-stage builds and one for the scripting update would be good if you're happy to create those?

twometresteve:
> It doesn't have to be part of this PR, but I think it'd definitely be a good time to consider whether the pipeline can be split into multi-stage builds at this point similar to how Android works - as the overall CI time is creeping up.
> 
> I'd be happy with just running tests on one Android/iOS/macOS platform when targeting an integration branch, and then running everything when targeting master/or when a manual trigger is pressed - this might need a bit of discussion with others though.

Yes, we've been putting some thought to this too.  We have a test epic coming up to enhance the way we do multi-stage builds (by dynamically generating pipelines), so it may make sense to wait for that pattern to establish itself before embarking on a change.  We also have the option of assigning another Unity license to speed the build up.

fractalwrench (approved):

PR #282, opened 2021-05-26
Set context correctly for Cocoa native layer

fractalwrench:
## Goal

Sets the context correctly for crashes which occur in the native layer of Cocoa. The current implementation was calling `config.context`, whereas it should be calling `[Bugsnag setContext]` - as this updates the information that is serialized into Cocoa crash reports.

## Testing

TODO: test on a physical device


rich-bugsnag:
@fractalwrench  Tested on iPhone and MacOS, works great!

rich-bugsnag (approved):

PR #283, opened 2021-05-27
chore(example) remove splash screen and make macOS IL2CPP default

rich-bugsnag:
## Goal

We test on IL2CPP more often and it kept reverting causing human errors when testing native crashes

## Changeset

Made .net 4 and IL2CPP default for builds

## Testing

Manual build tested

### DIFF
--- a/example/ProjectSettings/ProjectSettings.asset
+++ b/example/ProjectSettings/ProjectSettings.asset
@@ -3,10 +3,11 @@
 --- !u!129 &1
 PlayerSettings:
   m_ObjectHideFlags: 0
-  serializedVersion: 14
+  serializedVersion: 18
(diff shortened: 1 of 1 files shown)


fractalwrench (commented):
Has this been tested on Unity 2017? If it works then that's fine, but from just looking at the changes I'd be worried that it won't be able to recognize the new fields.

rich-bugsnag:
> Has this been tested on Unity 2017? If it works then that's fine, but from just looking at the changes I'd be worried that it won't be able to recognize the new fields.

@fractalwrench When running 2017 with macos on this branch it just defaults back to mono, but keeps .net 4. So there are no errors or issues imo.

kattrali (approved):

PR #284, opened 2021-05-27
Avoid using jcenter() in Unity gradle scripts

fractalwrench:
## Goal

Removes jcenter() from gradle build scripts in bugsnag-unity. Note that the bugsnag-android submodule does still contain jcenter() - this will be addressed in a separate ticket and then brought across as part of a bugsnag-android dependency update.

## Testing

Assembled each changed project, and relied on existing test coverage.


lemnik (approved):
LGTM

PR #289, opened 2021-05-31
fix (Unity 2021 builds): adapt post process build script to work with unity 2021

rich-bugsnag:
## Goal

Unity 2021 build were not completing properly because the post process script that adds the shell script to upload dsyms was assuming that a shell script phase already existed

## Design

I added a check to see if the phase exists, if not the i create one before running the rest of the script

## Changeset

I added a check to see if the phase exists, if not the i create one before running the rest of the script

## Testing

Manually tested and the xcode project is edited as expected.

In testing i found issues outside the scope of this fix that will be addressed separately


rich-bugsnag:
@fractalwrench @tomlongridge @kattrali @nickdowell 

Hi guys, while fixing this I found some Xcode issues that I haven’t been able to fix yet, perhaps it’s outside the scope of this bug fix and requires some discussion. 

The circumstances also suggest that perhaps we should not fix it and redesign it all together.

**The original issue:**

When building for iOS we have a post build script that modifies the generated Xcode project. One of the modifications it makes is to add a shell script phase. The script uploads the generated dsyms.

The issue was that the script assumed a shell script build phase already existed. In unity 2021 this is not the case and so it throws an exception.

I fixed this by adding a check for that phase, if one does not exist then I create one. This fix works and the builds are now completing without issue.

**The new issue:**

When testing the generated Xcode project I noticed that nothing was being uploaded. This is because since Xcode 10 you must also declare input files for the script. So I added this to the generated script.

Then the script found the files, but the upload hung at around 90% of the upload.

Uploading using curl in the terminal worked fine.

I tried getting Xcode to use a different version of curl but that also didn’t work.

**Next steps:**

We could try to fix this, but the situation raises an important question.

Due to the input file change in Xcode 11, this has not been working for quite some time, perhaps a year or longer and no one has noticed. 

So should we fix it or just take it out and consider a different approach?

src/BugsnagUnity/PostProcessBuild.cs
```diff
@@ -1,253 +1,80 @@
...
-              info.AdvanceCurrentLine();
+                ProcessBuildConfigurationSection(info);
+                info.AdvanceCurrentLine();
+            }            
+        }
+     
+        /// <summary>
+        /// Processes the build configuration section. Here we need to enable two
+        /// build flags relating to certain exception types. We also need to ensure
+        /// that the -ObjC linker flag has been applied.
```
kattrali, line 53:
```suggestion
        /// build flags relating to certain exception types.
```

kattrali (approved):

rich-bugsnag:
In the end we decided to remove the shell script and the part that adds the -ObjC linker flag as it's not needed anymore

PR #292, opened 2021-06-02
fix(log messages): adapt regex to ignore timestamps at the start of log messages

rich-bugsnag:
## Goal

The current regex was mistaking timestamps for the error class

## Design

A simple fix

## Changeset

Changed the regex that looks for an error class name

## Testing

Manually tested in editor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This version contains **breaking** changes, as bugsnag-unity has been updated to
 
 
 ### Bug fixes
+* Timestamps in log messages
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This version contains **breaking** changes, as bugsnag-unity has been updated to
 
 
 ### Bug fixes
+* Timestamps in log messages
```
kattrali, line 9: I'd give this a general description, for those without context. Maybe something like:

```suggestion
* Fix an issue where timestamps and other `:`-containing log message content was interpreted as the error class
```

kattrali (approved):

PR #293, opened 2021-06-02
style: add spaces to ANR detection label

kattrali:
## Goal

Previously the auto-generated labels would add spaces, this does the same


richardelms:
Oh well spotted!

rich-bugsnag (approved):
Well spotted!

PR #299, opened 2021-06-14
chore(code style): setup code format and set solution based rules for formatting

rich-bugsnag:
## Goal

The current formatting was highly unusual and hard to read, and was not set in the solution so different devs would all have different formatting when adding new code.

## Changeset

Changed all files and added a custom solution based style


kattrali (approved):
such style :lipstick:

PR #302, opened 2021-06-18
Remove duplicate native android events

kattrali:
## Goal

Avoid sending "duplicate" native crash reports, one from native detection and a second from the Unity error handler, which converts a native crash into either a log message (2017-2018.2) or an uncaught Java exception (2018.3+). This is an extension of #162.

As a part of this change, I made a few improvements to the unity path detection and error messaging in the Rakefile to make it clearer what's going on when builds fail to start.

## Design

Added a native callback to reject any events matching a native crash pattern, and updated the log message detector to check.

## Changeset

* `BugsnagUnity` class now exposes a `getNativeCallback` function, which returns on `OnErrorCallback` to perform event filtering. It might also be handy later, if we ever alter initialization to be done through a native plugin earlier in the launch cycle.
* As a result of the last point, the `bugsnag-android-unity` package now depends on `bugsnag-android-core` (and should be kept up to date when updating the `bugsnag-android` dependency).
* `Exception` class now verifies both the error class and message to detect C/C++ exceptions on Android and filter them out
* Updated the unit test for filtering to be more generic and not depend on the presence of `libbugsnag` (which was reliable in early versions but now not necessary).

## Testing

Added slowdowns to the native crash tests to ensure that if a second event is generated, it is detected and fails the test, in addition to ridiculous amounts of manual testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * Correct android session start times
   [#291](https://github.com/bugsnag/bugsnag-unity/pull/291)
 
+* Fix duplicate events being sent for Android C/C++ crashes
 
 ## 5.0.0 (2021-06-08)
 
--- a/Rakefile
+++ b/Rakefile
@@ -18,9 +18,7 @@ def unity_directory
   elsif ENV.has_key? "UNITY_DIR"
     ENV["UNITY_DIR"]
   else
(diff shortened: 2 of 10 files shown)


rich-bugsnag (approved):
Talked through it with Delisa in a zooooom call and all looks good to me 👯

PR #306, opened 2021-06-23
fix(ANR): allow other error classes to be reported

rich-bugsnag:
ANRs (and potentially other errors) where not reporting due to the duplicated events fix.

Changes:

Added a more specific check and allowed other error types to be reported

Testing: 

Tested for an ANR on device

### DIFF
--- a/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/BugsnagUnity.java
+++ b/bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/BugsnagUnity.java
@@ -25,12 +25,8 @@ public class BugsnagUnity {
                 Error error = event.getErrors().get(0);
                 String errorClass = error.getErrorClass();
                 String message = error.getErrorMessage();
-                if(discardedEventErrorClass.equals(errorClass))
-                {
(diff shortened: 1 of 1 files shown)


bugsnag-android-unity/src/main/java/com/bugsnag/android/unity/BugsnagUnity.java
```diff
@@ -25,8 +25,14 @@ public boolean onError(Event event) {
...
                 String message = error.getErrorMessage();
-                return discardedEventErrorClass.equals(errorClass) &&
-                    (message == null || !pattern.matcher(message).find());
+                if(discardedEventErrorClass.equals(errorClass))
+                {
+                    if(message == null || !pattern.matcher(message).find())
+                    {
+                        return false;
+                    }
+                }
```
kattrali, line 34:
```suggestion
                if(discardedEventErrorClass.equals(errorClass)) {
                    return message == null || !pattern.matcher(message).find();
                }
```

style nit: Java is typically same-line for brackets.

kattrali (approved):

fractalwrench (approved):

PR #309, opened 2021-06-24
build: remove unneeded versioning tasks

kattrali:
## Goal

Fix a potential issue in the packaging/release process where the version number of the library may be reported incorrectly due to timing differences between the time the branch was pushed (kicking off the build) and the time the tag was pushed (previously queried by the build task). This issue was surfaced by changes in the release process.

## Testing

Tested manually to ensure that the reported version is correct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.1.1 (2021-06-24)
+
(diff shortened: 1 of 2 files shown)


rich-bugsnag (approved):
Thanks for this fix! Looks good

PR #310, opened 2021-06-25
deps(bugsnag-cocoa): update bugsnag-cocoa to v6.9.7

nickdowell:


rich-bugsnag (approved):
Thanks for this @nickdowell

PR #317, opened 2021-06-30
chore(test) add test for automatic bugsnag loaded breadcrumb

rich-bugsnag:
## Goal

Add tests for the native breadcrumb 'Bugsnag loaded'

## Changeset

Added desktop, android and iOS tests to check for the existence on the specific breadcrumb

## Testing

Tested manually and against browserstack


twometresteve (approved):
LGTM

PR #319, opened 2021-06-30
Build Windows test fixtures

twometresteve:
## Goal

Build e2e test fixtures for Unity 2017 to 2020 on Windows in the CI pipeline.

## Design

New steps are currently behind a block step as we don't yet have permanent Windows build servers in place.  There's also low value in only building the test fixtures - running the e2e tests will come shortly.

To run these new build steps, a Buildkite agent can be set up temporarily on a Windows development machine very easily and the block step manually released through the Buildkite UI.

## Changeset

New pipeline steps as described.  Existing concurrency group settings removed as they are unnecessary - only the number of Unity agents for the number of licenses that we have will be stood up in the first place.  If there's an agent free, it will be licensed.

## Testing

Tested with an agent running on my Windows development machine.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -185,7 +185,6 @@ steps:
     command: sh -c .buildkite/pipeline_trigger.sh
 
   - block: "Build Windows fixtures"
-    key: block-windows
     prompt: "Build Windows fixtures - Is a suitable agent running?"
 
   #
@@ -238,7 +237,7 @@ steps:
 
   - label: Build Unity 2020 Windows test fixture
     key: 'windows-2020-fixture'
-#    depends_on: 'build-artifacts'
+    depends_on: 'build-artifacts'
     timeout_in_minutes: 30
     agents:
       queue: opensource-windows-unity
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -218,10 +218,10 @@ steps:
     command:
       # Using the artifacts plugin v1.3 on Windows seem to break the whole step
       - 'buildkite-agent artifact download "Bugsnag.unitypackage" .'
-      - 'cd test\\desktop'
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -191,3 +183,74 @@ steps:
...
+    timeout_in_minutes: 30
+    agents:
+      queue: opensource-windows-unity
+    env:
+      UNITY_VERSION: "2017.4.40f1"
+    command:
+      # Using the artifacts plugin v1.3 on Windows seem to break the whole step
+      - 'buildkite-agent artifact download "Bugsnag.unitypackage" .'
+      - 'cd test\\desktop'
+      - 'c:\\Progra~1\\Git\\bin\\bash.exe ./features/scripts/build_maze_runner.sh'
```
Cawllec, line 206: I expect the answer is no, but is there no way we can either avoid using both Win and Linux paths in a single line?

Likewise, is it possible to not need the escape slashes in the windows paths?

Cawllec (commented):
This looks good to me at the moment.  Are we looking at merging this down and then making changes when new test systems are provisioned for us, or will this be parked for a bit?

.buildkite/pipeline.yml
twometresteve: I think the escapes on `\` are needed because it's a YAML file, but now I think about it I think you can actually use forward slashes in Windows paths anyway... trialling that now.

twometresteve:
> This looks good to me at the moment. Are we looking at merging this down and then making changes when new test systems are provisioned for us, or will this be parked for a bit?

Yes, I'd merge it now so it doesn't go stale and update as soon as we get the infrastructure.  If engineers want they will also be able to set up a Buildkite agent on their dev machine easily and run the tests.

Cawllec (approved):
LGTM pending CI pass

PR #320, opened 2021-07-01
6923 6924 webgl ci setup

rich-bugsnag:
## Goal

Setup a fixture for WebGL CI testing

## Design

It was eary to convert the existing desktop fixture to parse the page url and take the test arguments from there

## Changeset

Added a way to parse the url for params in webgl builds
Added a webgl build script 

## Testing

Manually

### DIFF
--- a/test/desktop/features/scripts/build_maze_runner_web_gl.sh
+++ b/test/desktop/features/scripts/build_maze_runner_web_gl.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+if [ -z "$UNITY_VERSION" ]; then
+  echo "UNITY_VERSION must be set, to e.g. 2017.4.40f1"
+  exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


test/desktop/features/scripts/build_maze_runner_web_gl.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
twometresteve, line 1: At some point I'd look to unify this with the existing script, as there is a lot of commonality.  I wouldn't do it just yet, though, as I'm making some updates to it for Windows.  Indeed, I'll probably have cause to look at this script when I come to add the MazeRunner steps, so I could probably look at unifying it at that point.

twometresteve (approved):
LGTM

test/desktop/features/scripts/build_maze_runner_web_gl.sh
rich-bugsnag: in the end i think we would probably have one script that takes arguments. ./build_maze_runner --windows etc.

PR #327, opened 2021-07-08
task(breadcrumbs): add bugsnag loaded breadcrumb when using fallback

rich-bugsnag:
## Goal

The fallback was not leaving a bugsnag loaded breadcrumb on init

## Changeset

added a check for when the fallback is in use and left a crumb

## Testing

Tested manually and this is covered in existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@
 * Fixed an issue where Breadcrumbs were reported in the wrong order on Windows and in the Unity Editor
   [#322](https://github.com/bugsnag/bugsnag-unity/pull/322)
 
+* Fixed an issue where platforms using the fallback implementation were not leaving a Bugsnag loaded breadcrumb 
+  [#327](https://github.com/bugsnag/bugsnag-unity/pull/327)
 
 
 ## 5.1.1 (2021-06-24)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -76,6 +76,8 @@
 * Fixed an issue where Breadcrumbs were reported in the wrong order on Windows and in the Unity Editor
   [#322](https://github.com/bugsnag/bugsnag-unity/pull/322)
 
+* Fixed an issue where platforms using the fallback implementation were not leaving a Bugsnag loaded breadcrumb
```
kattrali, line 79: Since the behavior of the implementation is somewhat opaque, I'd list the conditions here where it might not have worked.

```suggestion
* Fixed an issue where a "Bugsnag loaded" breadcrumb was not added on Windows, Linux, WebGL, and editor builds
```

src/BugsnagUnity/Client.cs
```diff
@@ -95,6 +95,31 @@ public Client(INativeClient nativeClient)
...
+                Breadcrumbs.Leave("Bugsnag loaded", BreadcrumbType.State, null);
+            }
+        }
+
+        private bool IsUsingFallback()
+        {
+            switch (Application.platform)
+            {
+                case RuntimePlatform.OSXEditor:
+                case RuntimePlatform.OSXPlayer:
```
kattrali, line 114: Wouldn't `OSXPlayer` leave a duplicate? The `bugsnag-cocoa` integration should be passing this in automatically.

rich-bugsnag: ah yes good point

kattrali (approved):

PR #330, opened 2021-07-09
Resolve Windows e2e test failures

twometresteve:
## Goal

Resolves e2e test failures on Windows.

## Changeset

Cucumber steps adjusted to cater for differences between. macOS and Windows.  Some scenarios now run on macOS only as they don't apply to Windows.

## Testing

Now covered by CI, noting that at present we only have temporary test infrastructure for Windows, which needs to be set up manually.

### DIFF
--- a/src/BugsnagUnity/Native/Windows/NativeClient.cs
+++ b/src/BugsnagUnity/Native/Windows/NativeClient.cs
@@ -21,6 +21,7 @@ namespace BugsnagUnity
 
         public void PopulateApp(App app)
         {
+            app.AddToPayload("type", "Windows");
         }
 
         public void PopulateDevice(Device device)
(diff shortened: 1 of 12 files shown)


src/BugsnagUnity/Native/Windows/NativeClient.cs
```diff
@@ -21,6 +21,7 @@ public NativeClient(IConfiguration configuration)
 
         public void PopulateApp(App app)
         {
+            app.AddToPayload("type", "Windows");
```
fractalwrench, line 24: Has this been checked with other teams? My understanding is that this field can affect analytics reports and might require backend changes.

fractalwrench (approved):
LGTM pending resolution of one comment

src/BugsnagUnity/Native/Windows/NativeClient.cs
tomlongridge: Yep - this is the agreed string for Windows (as per Electron: https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-electron-app/app.js#L8)

PR #332, opened 2021-07-13
Fix e2e test failures for native errors on macOS

twometresteve:
## Goal

Fixes a few test failures that had gone unnoticed in the e2e tests, due to the return code not being fed back to Buildkite (also fixed).

## Design

Introduces a new Cucumber step to allow individual scenario steps to specify when the notifier name will be derived from the native notifier on macOS.

## Changeset

Also bumps the minor version on Unity 20202, after an incidental update of our build servers.

## Testing

Covered by a basic CI run.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -27,7 +27,7 @@ steps:
     agents:
       queue: opensource-mac-unity
     env:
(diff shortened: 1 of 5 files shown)


fractalwrench (approved):
LGTM pending CI

PR #339, opened 2021-07-22
task(ANRs): make ANR detection enabled as default

rich-bugsnag:
## Goal

Product have requested that this is now enabled by default and approved this under a minor release.

## Changeset

Made the default in the config class and the bugsnag behaviour class true

## Testing

Manually tested


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Enhancements
 
+* Automatic ANR detection is now enabled by default
```
kattrali, line 7: Since this is a platform-specific thing, I try to over-explain / link to related docs for stuff like this:

```suggestion
* [Android] Automatic [App Not Responding](https://developer.android.com/topic/performance/vitals/anr) (ANR) detection is now enabled by default  
```

kattrali (approved):

PR #340, opened 2021-07-23
Merge mobile into desktop e2e test fixture

twometresteve:
## Goal

Merge the mobile test fixture and tests in `test/mobile` into `features`.  There is now just one, universal e2e test fixture, although the mobile tests will rename separate for the time being.  They will be unified with the "main" features in due course.

## Design

The test fixture has been updated to included numbers 0-9 and a "dial" button, which are only used by the mobile tests.  It replaces the fixed number of buttons (each for a specific scenario), allowing us to scale on the number of features (and, in particular, eventually have just one set of features for all Unity platforms).

I'm sure there are elements of the setup here that could be made more elegant.  For example, I think we could have a single lookup table for the scenario numbers/names in a separate text file.  This would allow both the test fixture and Cucumber steps to read from a single point of truth. 

## Changeset

- Several files moved verbatim (this PR isn't as bad as it first looks)
- Test fixtures and Cucumber step updates to "dial up" test scenarios by number
- Build script and pipeline changes for the new file path
- Wording of mobile scenario steps altered to avoid clashing with the existing "desktop" tests

## Testing

Covered by a full CI run.


features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
```diff
@@ -54,24 +55,144 @@ public void TestMaxBreadcrumbs()
...
+        var input = ScenarioName.text;
+
+        if (string.IsNullOrEmpty(input) || !input.EndsWith("#"))
+        {
+            return;
+        }
+
+        var scenarioName = input.Remove(input.Length - 1);
+
+        switch (scenarioName)
```
fractalwrench, line 148: Would be good to add a `default` branch here that throws an exception stating no scenario could be found with that name.

features/steps/unity_steps.rb
```diff
@@ -1,3 +1,81 @@
...
+  }
+  number = lookup[scenario]
+  $logger.debug "#{scenario}' has dial-in code #{number}"
+
+  step("I dial #{number / 10}")
+  sleep 1
+  step("I dial #{number % 10}")
+  sleep 1
+  step("I press dial")
+  sleep 1
```
fractalwrench, line 40: Are these sleeps necessary for the test to work?

fractalwrench (approved):
LGTM once comments are resolved and a full CI run has taken place.

features/steps/unity_steps.rb
twometresteve: I put them in for now just to help gain some confidence in the pressing of the numbered buttons - so that we can see them being entered on the captured video.  It's possibly not needed and certainly one to review over time, although I'm not too concerned given the number of scenarios we currently have in place.

features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
twometresteve: 👍

rich-bugsnag (approved):
Thanks Steve!

PR #341, opened 2021-07-26
6675 enabled error types

rich-bugsnag:
## Goal

New configuration option EnabledErrorTypes which can be set with a set of an enumerated type of the following values: https://github.com/bugsnag/platforms-notifier-spec/blob/master/event.md#enablederrortypes 

## Changeset

- Added enum selector for the bugsnag gameobject 
- Added fallback implementation of the ErrorType class and detection to filter error types in the client script
- Passed the value to Android and Cocoa
## Testing

Tested manually and added tests


example/Logs/shadercompiler-UnityShaderCompiler3.log
```diff
@@ -0,0 +1,210 @@
+Base path: '/Applications/Unity/Hub/Editor/2020.3.7f1/Unity.app/Contents', plugins path '/Applications/Unity/Hub/Editor/2020.3.7f1/PlaybackEngines'
```
kattrali, line 1: Assuming adding this file was unintended ...

src/BugsnagUnity/ErrorTypes.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+namespace BugsnagUnity
+{
+    public class ErrorTypes
```
kattrali, line 4: More of a general question but do you feel that the usage of `ErrorTypes` is idiomatic (for C# in general and Unity specifically)? The combination of setting various fields to `true` and/or using `SetAllDisabled()` seems a bit convoluted for a property on configuration though I haven't seen a lot of similar implementations. Would a flag enum be too low level?

kattrali (approved):

PR #343, opened 2021-07-28
refac (mobile tests): refactor mobile tests to enable custom configs and faster run times

rich-bugsnag:
## Goal

It was not possible to easily set custom config options in the unity mobile test fixture, and tests were all taking 8 seconds longer than needed

## Changeset

Changed the way the MobileTestRunner class is organised 

## Testing

Full manual and CI run passing


twometresteve (approved):
LGTM

PR #345, opened 2021-08-02
Soft fail on Android for Unity 2017

twometresteve:
## Goal

Soft fail on Android for Unity 2017 pending further investigation of missing field `metaData.device.osLanguage` after a recent change to the test fixture (it is present when tested manually in other apps).


rich-bugsnag (approved):
Thanks @twometresteve

PR #347, opened 2021-08-02
task(config): new configuration option AppHangThresholdMillis

rich-bugsnag:
## Goal

New configuration option AppHangThresholdMillis to set the native Cocoa option: https://docs.bugsnag.com/platforms/ios/configuration-options/#apphangthresholdmillis

Cocoa-only.

## Testing

Manually tested, no ci tests planned due to difficulty automating the apphang

### DIFF
--- a/src/BugsnagUnity/Configuration.cs
+++ b/src/BugsnagUnity/Configuration.cs
@@ -112,6 +112,8 @@ namespace BugsnagUnity
 
         public ErrorTypes[] EnabledErrorTypes { get; set; }
 
+        public int AppHangThresholdMillis { get; set; } = 5000;
+
         public virtual bool IsErrorTypeEnabled(ErrorTypes errorType)
         {
             return EnabledErrorTypes == null || EnabledErrorTypes.Contains(errorType);
--- a/src/BugsnagUnity/Configuration.cs
+++ b/src/BugsnagUnity/Configuration.cs
@@ -66,9 +66,10 @@ namespace BugsnagUnity
             {
                 if (value < 0 || value > 100)
                 {
-#if UNITY_EDITOR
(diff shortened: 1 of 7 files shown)


src/BugsnagUnity/Configuration.cs
```diff
@@ -138,6 +138,8 @@ public virtual bool AutoTrackSessions
 
         public ErrorTypes[] EnabledErrorTypes { get; set; }
 
+        public int AppHangThresholdMillis { get; set; } = -1;
```
kattrali, line 141: This should probably be `ulong`, to match the precision of [`NSUInteger`](https://developer.apple.com/documentation/objectivec/nsuinteger).

kattrali (commented):
The logic looks good :+1: - added a minor concern over integer types

src/BugsnagUnity/Configuration.cs
rich-bugsnag: ulong can't be -1, how about we split the difference and go for a long?

rich-bugsnag: perhaps i sould also add a limit like MaximumBreadcrumbs

kattrali: The [threshold on iOS must be > 250](https://github.com/bugsnag/bugsnag-cocoa/blob/b83847c74669f64a9afc0e7e7ade595b620c09e8/Bugsnag/include/Bugsnag/BugsnagConfiguration.h#L210-L220), so maybe zero could be the sentinel value?

rich-bugsnag: Done 👍

kattrali (approved):

PR #348, opened 2021-08-02
task(tests): add ability to clear persistent data in iOS test fixture

rich-bugsnag:
add ability to clear persistent data in iOS test fixture


twometresteve (approved):
There's some excess whitespace around the new method, but it's not worth waiting for another build for - and I can sort it out when plumbing the new functionality into the tests.

PR #349, opened 2021-08-03
refac(API): change notify level to notifyloglevel

rich-bugsnag:
## Goal

Notifier Spec Compliance

### DIFF
--- a/src/BugsnagUnity/IConfiguration.cs
+++ b/src/BugsnagUnity/IConfiguration.cs
@@ -49,7 +49,7 @@ namespace BugsnagUnity
 
         string Context { get; set; }
 
-        LogType NotifyLevel { get; set; }
+        LogType NotifyLogLevel { get; set; }
         
         bool AutoDetectErrors { get; set; }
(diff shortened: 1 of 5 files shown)


src/BugsnagUnity/IConfiguration.cs
```diff
@@ -49,7 +49,7 @@ public interface IConfiguration
 
         string Context { get; set; }
 
-        LogType NotifyLevel { get; set; }
+        LogType NotifyLogLevel { get; set; }
```
kattrali, line 52: Is this a breaking change for anyone initializing bugsnag-unity via code rather than game object? I would have imagined both would need to be defined on the interface for `NotifyLevel` to remain accessible.

rich-bugsnag: NotifyLevel is still there in the public configuration class (and has a warning on it now). 

The interface is only referenced behind the scenes in the client etc

kattrali (approved):

PR #350, opened 2021-08-03
fix(stacktraces): add support for windows new line characters

rich-bugsnag:
## Goal

Windows stacktraces were not being split properly

## Design

Added support for more types of new line

## Testing

Manually tested and fix a test that should have caught this issue before

### DIFF
--- a/src/BugsnagUnity/Payload/StackTraceLine.cs
+++ b/src/BugsnagUnity/Payload/StackTraceLine.cs
@@ -31,7 +31,7 @@ namespace BugsnagUnity.Payload
 
         internal StackTrace(string stackTrace, StackTraceFormat format)
         {
-            string[] lines = stackTrace.Split(new[] { System.Environment.NewLine },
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/Payload/StackTraceLine.cs
```diff
@@ -31,7 +31,7 @@ internal StackTrace(StackFrame[] stackFrames)
 
         internal StackTrace(string stackTrace, StackTraceFormat format)
         {
-            string[] lines = stackTrace.Split(new[] { System.Environment.NewLine },
+            string[] lines = stackTrace.Split(new[] {"\r\n","\r","\n", System.Environment.NewLine },
```
twometresteve, line 34: Is this guaranteed to split in precedence order of the array elements?  Or could a `\r\n` sequence end up being split twice, resulting in a blank entry?

twometresteve (approved):
LGTM - one question asked just to be sure, but I suspect it's correct as it is.

PR #355, opened 2021-08-05
Bump bugsnag-android to v5.11.0

fractalwrench:
## Goal

Bumps the bugsnag-android dependency to v5.11.0.


rich-bugsnag (approved):
Thanks @fractalwrench

PR #358, opened 2021-08-06
Avoid Appium session restarts between iOS e2e scenarios

twometresteve:
## Goal

Avoid have to use the `--separate-sessions` MazeRunner option to create a new Appium session before each iOS test scenario.  This speeds up the run and also means that we just get a single video and set of logs for the run to inspect.

## Design

I've reworked the test fixture a little to hook into the existing functionality to clear all persistent data on iOS.  It was the "false OOMs" caused by the app being killed between scenarios that meant we needed to use separate sessions initially.

I've separated the running of commands from scenarios.  Although they are very similar in how they are invoked, they are separate concerns and have some important differences.  For example, command should not typically be starting Bugsnag.

## Changeset

- `--separate-sessions` removed from pipelines
- Test fixture buttons reduced in height slightly to fit a new "Run command" button on.
- Test fixture and harness changes to separate the concerns of running scenarios and commands, whilst reusing common elements.

## Testing

Covered by a full CI run.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -553,7 +553,6 @@ steps:
   #
   # Run iOS tests
   #
-  # TODO: PLAT-6656 - Avoid having to use --separate-sessions
   - label: ':ios: Run iOS e2e tests for Unity 2017'
     depends_on: 'build-ios-fixture-2017'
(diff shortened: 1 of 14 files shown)


rich-bugsnag (approved):
Good idea, thanks @twometresteve

PR #359, opened 2021-08-06
task(config): add new BundleVersion configuration option

rich-bugsnag:
## Goal

New configuration option BundleVersion to set the native Cocoa bundleVersion option

## Design

The config value is null by default and only passed through to cocoa if set by the user

## Changeset

- Added BundleVersion to IConfiguration and Configuration classes
- Added native method bugsnag_setBundleVersion in NativeCode and Bugsnag.mm classes

## Testing

Added check for the value in the iOS test ios_csharp_errors.feature:8

### DIFF
--- a/src/BugsnagUnity/Native/Cocoa/NativeClient.cs
+++ b/src/BugsnagUnity/Native/Cocoa/NativeClient.cs
@@ -34,6 +34,10 @@ namespace BugsnagUnity
             NativeCode.bugsnag_setAppVersion(obj, config.AppVersion);
             NativeCode.bugsnag_setNotifyUrl(obj, config.Endpoint.ToString());
             NativeCode.bugsnag_setMaxBreadcrumbs(obj, config.MaximumBreadcrumbs);
+            if (!string.IsNullOrEmpty(config.BundleVersion))
+            {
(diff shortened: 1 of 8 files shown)


nickdowell (approved):
LGTM

src/BugsnagUnity/Native/Cocoa/NativeClient.cs
```diff
@@ -34,6 +34,10 @@ IntPtr CreateNativeConfig(IConfiguration config)
             NativeCode.bugsnag_setAppVersion(obj, config.AppVersion);
             NativeCode.bugsnag_setNotifyUrl(obj, config.Endpoint.ToString());
             NativeCode.bugsnag_setMaxBreadcrumbs(obj, config.MaximumBreadcrumbs);
+            if (!string.IsNullOrEmpty(config.BundleVersion))
```
nickdowell, line 37: This if statement doesn't seem to be necessary - we don't check `IsNullOrEmpty` for `AppVersion` or `ReleaseStage`

twometresteve: The (internal) spec says:
> The `bundleVersion` property shall be set from supplied `Configuration`.

So yes, I agree - if it's null/empty then it should be set as such in the native layer.

PR #360, opened 2021-08-08
Resolve e2e test flake with open

twometresteve:
## Goal

Resolve a curious test flake that occurs in the Buildkite CI environment when using `open` to launch the test fixture.

## Design

I was unable to explain the reason that `open` caused a test flake, but invoking the executable directly instead seems to have resolved it.

## Testing

Tested by running macOS e2e tests on all Unity versions on an hour schedule for a day.


rich-bugsnag (approved):
Thanks @twometresteve

PR #363, opened 2021-08-12
task(config): add new config option AppType and pass through to android and ios

rich-bugsnag:
## Goal

add new config option AppType and pass through to android and ios

## Changeset

- Added AppType to C# Config class
- Passed the value to android and iOS
- Added windows native implementation to PopulateApp method
- Updated Fallback PopulateApp method

## Testing

Manually tested and added Desktop, Android and iOS CI tests

### DIFF
--- a/features/desktop/handled_errors.feature
+++ b/features/desktop/handled_errors.feature
@@ -32,6 +32,14 @@ Feature: Handled Errors and Exceptions
         And the first significant stack frame methods and files should match:
             | Main.DoNotify()           |
 
+    Scenario: Custom App Type
+        When I run the game in the "CustomAppType" state
+        And I wait to receive an error
+        Then the error is valid for the error reporting API sent by the Unity notifier
+        And the event "app.type" equals "test"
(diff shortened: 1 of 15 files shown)


features/desktop/handled_errors.feature
```diff
@@ -32,6 +32,14 @@ Feature: Handled Errors and Exceptions
         And the first significant stack frame methods and files should match:
             | Main.DoNotify()           |
 
+    Scenario: Custom App Type
```
twometresteve, line 35: Projecting forwards, I presume you will have lots of custom config tests that you'll want to run.  Why not just call this "Custom Config" and add fields to it over time?   Then you can check them all in one go and save a load of test run time.

twometresteve: (or you could leave it as-is and modify for the next field that's added)

rich-bugsnag: Great idea, will change it for the next one 👍

nickdowell (approved):

PR #364, opened 2021-08-12
task(config): add android pass through config option ProjectPackages

rich-bugsnag:
## Goal

add android pass through config option ProjectPackages

## Changeset

- Added ProjectPackages to config class
- Pass the value through to android on init

## Testing

Manually tested, CI tests also added

### DIFF
--- a/features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
@@ -175,8 +175,6 @@ public class MobileScenarioRunner : MonoBehaviour {
                 LogError();
                 break;
             case "Java Background Crash":
-                LeaveBreadcrumbString();
-                LeaveBreadcrumbTuple();
                 MobileNative.TriggerBackgroundJavaCrash();
                 break;
             case "Native exception":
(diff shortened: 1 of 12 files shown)


twometresteve:
Why aren't CI tests possible?  I can see that we can't check the value set in the Android layer directly, but can't we infer that it's been set correctly by having a scenario that:
- Sets a single ProjectPackage
- Causes a native error
- Verifies that a stack frame is marked as inProject and another not

If the ProjectPackage is set differently to what the default would be, that should be sufficient to infer that the config has taken effect.

rich-bugsnag:
@twometresteve Added tests 👍

features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
```diff
@@ -168,6 +174,11 @@ public void DoTestAction(string scenarioName)
                 SetUser();
                 LogError();
                 break;
+            case "Java Background Crash":
+                LeaveBreadcrumbString();
+                LeaveBreadcrumbTuple();
```
lemnik, line 179: Given the separation of "throw Exception" and "throw Exception with breadcrumbs": should "Java Background Crash" really be leaving breadcrumbs before crashing?

rich-bugsnag: Removed the unnecessary calls to the breadcrumb methods

lemnik (approved):
LGTM

PR #366, opened 2021-08-16
task(config): add configuration option Endpoints

rich-bugsnag:
## Goal

-New configuration option Endpoints to set the endpoints (via a new class EndpointConfiguration)

-The existing Endpoint and SessionEndpoint configuration options should be marked as deprecated and alias the new option.

## Changeset

- Created new EndpointConfiguration.cs class and added the new field Endpoints to the config class
- Used the build in System.Uri class to validate the input of the constructor.
- Added a check to session tracking to not send sessions if the config is not valid and log a warning.
- Added a check to the client class to not send error events if the config is not valid, did not add a warning as this could lead to a loop if the notify level is configured to send warning logs as events.
- Added deprecation warnings for config.Endpoint and config.SessionEndpoint
- Made config.Endpoint and config.SessionEndpoint alias the new Endpoints class fields

## Testing

Added unit tests to check for config validation

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -260,7 +260,7 @@ namespace BugsnagUnity
 
         void Notify(Exception[] exceptions, HandledState handledState, Middleware callback, LogType? logType)
         {
-            if (!ShouldSendRequests() || EventContainsDiscardedClass(exceptions))
+            if (!ShouldSendRequests() || EventContainsDiscardedClass(exceptions) || !Configuration.Endpoints.IsValid)
             {
                 return; // Skip overhead of computing payload to to ultimately not be sent
             }
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -95,7 +95,25 @@ namespace BugsnagUnity
             {
                 // Async behavior is not available in a test environment
             }
+            if (!Configuration.Endpoints.IsValid)
+            {
+                CheckForMisconfiguredEndpointsWarning();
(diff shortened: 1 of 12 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -95,9 +95,27 @@ public Client(INativeClient nativeClient)
...
+        {
+            var endpoints = Configuration.Endpoints;
+            if (endpoints.NotifyIsCustom && !endpoints.SessionIsCustom)
+            {
+                UnityEngine.Debug.LogWarning("Invalid configuration. endpoints.Notify cannot be set without also setting endpoints.Session. Events will not be sent to Bugsnag.");
+            }
+            if (!endpoints.NotifyIsCustom && endpoints.SessionIsCustom)
+            {
+                UnityEngine.Debug.LogWarning("Invalid configuration. endpoints.Session cannot be set without also setting endpoints.Notify. Sessions will not be sent to Bugsnag.");
+            }
```
nickdowell, line 115: Not sure about this logic. The spec says "If `endpoints.notify` is provided by the developer, the notifier shall validate that `endpoints.session` is also provided" - but it's ambiguous whether that "provided" means "is different from the default"... @kattrali ?

rich-bugsnag: The spec states that this is to avoid "leaking data to bugsnag", I checked with Delisa and she said the idea is that we must never allow one type of endpoint to be default if the other is not. So i think this logic is correct.

nickdowell (approved):

PR #367, opened 2021-08-17
task(config): new configuration option RedactedKeys

rich-bugsnag:
## Goal

Setup metadata key redaction for the fallback/Windows and send the value through to iOS and Android

## Changeset

- Added RedactedKeys array to the config class
- Check for redacted keys in the client before serialising an event (as specificed by the notifier spec)
- Pass the value through to android and iOS

## Testing

Manually and added CI tests

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -284,6 +284,8 @@ namespace BugsnagUnity
                 metadata.AddToPayload(item.Key, item.Value);
             }
 
+            CheckMetadaForRedactedKeys(metadata);
+
             var @event = new Payload.Event(
               Configuration.Context,
               metadata,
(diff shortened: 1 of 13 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -353,6 +355,36 @@ void Notify(Exception[] exceptions, HandledState handledState, Middleware callba
             }
         }
 
+        private void CheckMetadaForRedactedKeys(Metadata metadata)
```
nickdowell, line 358: The name `CheckMetadaForRedactedKeys` doesn't make it clear that this function will actually redact the metadata. Maybe `RedactMetadata` would be a clearer?

nickdowell (approved):

src/BugsnagUnity/Client.cs
rich-bugsnag: agreed, thanks 👍

PR #368, opened 2021-08-17
task(config): add new config option PersistenceDirectory

rich-bugsnag:
## Goal

New configuration option PersistenceDirectory to set the native Android `persistenceDirectory`: https://docs.bugsnag.com/platforms/android/configuration-options/#persistencedirectory.

## Changeset

- added PersistenceDirectory to config class
- Pass the value through to andorid

## Testing

Manually tested and added CI test


lemnik (approved):
LGTM

features/android/android_custom_configs.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Android custom config setting
+
+    Background:
+        Given I wait for the mobile game to start
+
+    Scenario: Android Persistence Directory
+        When I run the "Android Persistence Directory" mobile scenario
+        Then I wait to receive an error
```
twometresteve, line 8: I would have added a check on the contents of the exception to ensure it's the exception we expect.  E.g. include a message when throwing it to say "Persistence directory is where I expect it".  Otherwise the test could fall over in a heap and we would be none the wiser.

rich-bugsnag: great idea, will add it in a new pr

PR #372, opened 2021-08-18
task(config): new configuration option PersistUser

rich-bugsnag:
## Goal

New configuration option PersistUser to set the fallback, native Cocoa and android values

## Changeset

- Added  to the config class and interface
- Pass the value through to Cocoa and Android

## Testing

Manually tested, full CI tests will be added when unity persistance layer is added

### DIFF
--- a/src/BugsnagUnity/Native/Cocoa/NativeCode.cs
+++ b/src/BugsnagUnity/Native/Cocoa/NativeCode.cs
@@ -43,7 +43,10 @@ namespace BugsnagUnity
         internal static extern void bugsnag_setAutoNotify(bool autoNotify);
 
         [DllImport(Import)]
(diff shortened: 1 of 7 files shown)


src/BugsnagUnity/Native/Cocoa/NativeCode.cs
```diff
@@ -43,7 +43,10 @@ partial class NativeCode
         internal static extern void bugsnag_setAutoNotify(bool autoNotify);
 
         [DllImport(Import)]
-        internal static extern void bugsnag_setAutoNotifyConfig(IntPtr configuration, bool autoNotify);       
+        internal static extern void bugsnag_setAutoNotifyConfig(IntPtr configuration, bool autoNotify);
```
nickdowell, line 46: Can't see what's changed here 😅

PR #380, opened 2021-08-30
task(api): add notify variants with stacktrace parameter

rich-bugsnag:
## Goal

add notify variants with stacktrace parameter

## Changeset

Added the following notify overloads:

Notify(exception, stacktrace) 

Notify(exception, stacktrace, callback)

Notify(name, message, stacktrace)

Notify(name, message, stacktrace, callback) 

## Testing

Added CI tests

### DIFF
--- a/features/desktop/handled_errors.feature
+++ b/features/desktop/handled_errors.feature
@@ -13,6 +13,28 @@ Feature: Handled Errors and Exceptions
             | Main.RunScenario(string scenario)  |
             | Main.Start()        |
 
+    Scenario: Calling notify with a custom stacktrace
+        When I run the game in the "CustomStacktrace" state
+        And I wait to receive an error
+        Then the error is valid for the error reporting API sent by the Unity notifier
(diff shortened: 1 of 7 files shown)


features/desktop/handled_errors.feature
```diff
@@ -13,6 +13,28 @@ Feature: Handled Errors and Exceptions
...
+        And the first significant stack frame methods and files should match:
+            | Main.CUSTOM()      |
+            | Main.CUSTOM()  |
+
+    Scenario: Calling notify with only strings
+        When I run the game in the "NotifyWithStrings" state
+        And I wait to receive an error
+        Then the error is valid for the error reporting API sent by the Unity notifier
+        And the exception "errorClass" equals "CUSTOM"
+        And the exception "message" equals "CUSTOM"
```
fractalwrench, line 32: Could the values be different for `errorClass/message` to confirm that the right value is definitely being populated?

fractalwrench (approved):

PR #382, opened 2021-09-01
task(sessions): move session tracking to fallback only

rich-bugsnag:
## Goal

Unity layer session tracking logic will be removed in favor of using the functionality provided by native layer(s). 

## Changeset

- Added the methods 'StartSession, ResumeSession, PauseSession, GetCurrentSession and UpdateSession' to the Native Client interface
- Setup Cocoa and Android bridging code for the aforementioned methods
- 'GetCurrentSession' is now used by csharp layer events to get the session info from the native layer or fallback
- 'UpdateSession' is used to make sure the session in the native layer is made aware of exceptions reported by the csharp layer, thus keeping the handled/unhandled counts in sync
- Changed fallback session tracking to only track sessions when fallback is used
- Sent value of Config.AutoTrackSessions to Cocoa and Android

## Testing

Manually tested and updated all CI tests to handle the new session payloads.


rich-bugsnag:
Before merging the CI tests for this will need to be revisited after the following tasks are complete:

https://bugsnag.atlassian.net/browse/PLAT-7224

https://bugsnag.atlassian.net/browse/PLAT-7190

features/android/android_sessions.feature
```diff
@@ -8,7 +8,7 @@ Feature: Android smoke tests for sessions
         And I wait to receive a session
 
         # Session payload
-        And the session payload field "notifier.name" equals "Unity Bugsnag Notifier"
+        And the session payload field "notifier.name" equals "Bugsnag Unity (Android)"
```
nickdowell, line 11: 🤨

src/BugsnagUnity/Client.cs
```diff
@@ -429,7 +429,17 @@ public void SetApplicationState(bool inFocus)
...
+                        if (IsUsingFallback())
+                        {
+                            SessionTracking.StartSession();
+                        }
+                        else
+                        {
+                            if (Application.platform.Equals(RuntimePlatform.Android))
+                            {
+                                SessionTracking.ResumeSession();
+                            }
```
nickdowell, line 441: Why only Android?

rich-bugsnag: Because Android cannot listen to the activity lifecycle whereas Cocoa does right?

features/ios/ios_sessions.feature
```diff
@@ -51,6 +45,6 @@ Feature: iOS smoke tests for sessions
         And the error payload field "events.0.user.id" equals the stored value "automated_user_id"
         And the error payload field "events.0.device.id" equals the stored value "automated_user_id"
         And the event "session.id" is not null
-        And the error payload field "events.0.session.id" equals the stored value "automated_session_id"
+        #And the error payload field "events.0.session.id" equals the stored value "automated_session_id"
```
fractalwrench, line 48: Are these not equal anymore?

src/BugsnagUnity/Native/Android/NativeClient.cs
```diff
@@ -20,6 +20,10 @@ public NativeClient(IConfiguration configuration)
             Configuration = configuration;
             Delivery = new Delivery();
             Breadcrumbs = new Breadcrumbs(NativeInterface);
+            if (configuration.AutoTrackSessions)
```
fractalwrench, line 23: This should preferably be `ResumeSession()` to avoid the edge case where a session has already been automatically started from an activity lifecycle callback. Also it looks like `Client.cs` might already be doing this?

src/BugsnagUnity/Payload/Report.cs
```diff
@@ -84,6 +84,12 @@ public void Ignore()
         /// <value>The app.</value>
         public App App => Event.App;
 
+        /// <summary>
+        /// Gets the session data that was gathered for this report.
+        /// </summary>
+        /// <value>The session.</value>
+        public Session Session => Event.Session;
```
fractalwrench, line 91: The spec seems to state that this shouldn't be publicly accessible, and Android also doesn't expose it

features/android/android_sessions.feature
rich-bugsnag: fixed 👍

src/BugsnagUnity/Payload/Event.cs
```diff
@@ -34,6 +33,8 @@ internal Event(string context, Metadata metadata, App app, Device device, User u
                     session.Events.IncrementUnhandledCount();
                 }
             }
+            this.AddToPayload("session", session);
```
fractalwrench, line 36: `IncrementHandledCount` and `IncrementUnhandledCount` don't look like they will increment the handled/unhandled count for the Android/Cocoa layer - will that cause issues?

rich-bugsnag: After the event is sent the Client calls UpdateSession in the NativeClient which updates the handled/unhandled values in the native layers.

src/BugsnagUnity/Payload/Report.cs
rich-bugsnag: Fixed 👍

features/ios/ios_sessions.feature
rich-bugsnag: Fixed, i forgot to switch it to use the new ignore case comparison step 👍

src/BugsnagUnity/Native/Android/NativeClient.cs
rich-bugsnag: Changed to Resume 👍 
The ResumeSession call in the client only happens when if the app has been open for more than 30 seconds and it gets focus again.

src/BugsnagUnity/Client.cs
nickdowell: Ahh 💡 not obvious - maybe worthy of a comment?

nickdowell (approved):

src/BugsnagUnity/Client.cs
rich-bugsnag: Done 👍

features/desktop/sessions.feature
```diff
@@ -4,12 +4,9 @@ Feature: Session Tracking
     Scenario Outline: Automatically receiving a session
         When I run the game in the "<scenario>" state
         And I wait to receive a session
-        Then the session is valid for the session reporting API version "1.0" for the "Unity Bugsnag Notifier" notifier
+        Then the session is valid for the session reporting API version "1.0" for the "OSX Bugsnag Notifier" notifier
         And the session payload field "app.version" is not null
         And the session payload field "app.releaseStage" equals "production"
-        And the session payload field "app.type" equals the platform-dependent string:
```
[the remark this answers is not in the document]
rich-bugsnag: Added a method to set the default app.type if non is set and re added the tests for it 👍

PR #388, opened 2021-09-10
[PLAT-7190] Set notifier name before starting Cocoa notifier

nickdowell:
## Goal

Ensure payloads sent during `Bugsnag.start()` have "Unity Bugsnag Notifier" in `notifier.name`.

## Changeset

`bugsnag-cocoa` now allows the `notifier` payload to be set on the configuration object (via a private API) before start. This change has not been included in a release yet, so this PR uses the `next` branch of `bugsnag-cocoa`.
* https://github.com/bugsnag/bugsnag-cocoa/pull/1182

This PR makes use of the new mechanism to configure the notifier info before start.

The notifier name has been changed from `"Bugsnag Unity (Cocoa)"` to `"Unity Bugsnag Notifier"` after discussion with @tomlongridge 

## Testing

Updated E2E test steps to verify that the notifier name is as expected.

### DIFF
--- a/features/steps/unity_steps.rb
+++ b/features/steps/unity_steps.rb
@@ -166,7 +166,7 @@ end
 
 Then("the error is valid for the error reporting API sent by the native Unity notifier") do
   # This step currently only applies to native errors on macOS macOS
-  check_error_reporting_api 'OSX Bugsnag Notifier'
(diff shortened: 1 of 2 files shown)


rich-bugsnag (approved):
Thanks @nickdowell !

PR #390, opened 2021-09-20
task(metadata): add app and device metadata fields to fallback/windows notifier

rich-bugsnag:
## Goal

Add app and device metadata fields to fallback/windows notifier

## Changeset

- Added generic (data present on both windows and fallbacks) data collection to the App and Device constructors
- Added platform specific data collection (lowmemory, isLaunching) to the PopulateApp and PopulateDevice methods in the fallback and windows NativeClient classes
- Added a lowMemoryWarning listener to the fallback and windows native client classes
- Added MarkLaunchAsCompleted implementation to the fallback and windows native client classes

## Testing

Added Windows CI test

### DIFF
--- a/features/desktop/unhandled_errors.feature
+++ b/features/desktop/unhandled_errors.feature
@@ -29,9 +29,9 @@ Feature: Reporting unhandled events
         
         #device metadata
         And the event "device.charging" is not null
-        And the event "device.batteryLevel" is not null
-        And the event "device.freeDisk" is not null
-        And the event "device.freeMemory" is not null
+        And the event "device.batteryLevel" is greater than -1
+        And the event "device.freeDisk" is greater than 0
+        And the event "device.freeMemory" is greater than 0
         And the event "device.hostname" is not null
         And the event "device.id" is not null
         And the event "device.locale" is not null
@@ -40,15 +40,15 @@ Feature: Reporting unhandled events
         And the event "device.osName" is not null
         And the event "device.osVersion" is not null
         And the event "device.runtimeVersions" is not null
-        And the event "device.screenDensity" is not null
+        And the event "device.screenDensity" is greater than 0
         And the event "device.screenResolution" is not null
         And the event "device.time" is a timestamp
         And the event "device.timezone" is not null
-        And the event "device.totalMemory" is not null
+        And the event "device.totalMemory" is greater than 0
 
         #app metadata
-        And the event "app.duration" is not null
-        And the event "app.durationInForeground" is not null
+        And the event "app.duration" is greater than 0
+        And the event "app.durationInForeground" is greater than 0
         And the event "app.inForeground" is not null
         And the event "app.isLaunching" is not null
         And the event "app.lowMemory" is not null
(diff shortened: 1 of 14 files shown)


features/desktop/unhandled_errors.feature
```diff
@@ -13,6 +13,50 @@ Feature: Reporting unhandled events
...
+        And the event "device.id" is not null
+        And the event "device.locale" is not null
+        And the event "device.manufacturer" is not null
+        And the event "device.model" is not null
+        And the event "device.osName" is not null
+        And the event "device.osVersion" is not null
+        And the event "device.runtimeVersions" is not null
+        And the event "device.screenDensity" is not null
+        And the event "device.screenResolution" is not null
+        And the event "device.time" is not null
```
fractalwrench, line 45: Would it be possible to validate that this is a timestamp, and that other app/device fields are non-zero integers etc?

src/BugsnagUnity/Native/Fallback/NativeClient.cs
```diff
@@ -12,18 +13,42 @@ class NativeClient : INativeClient
 
         public IDelivery Delivery { get; }
 
-        private Dictionary<string, object> _editorMetadata = new Dictionary<string, object>();
+        private Dictionary<string, object> _fallbackMetadata = new Dictionary<string, object>();
+
+        private bool _launchMarkedAsCompleted = false;
+
+        private bool _hasRecivedLowMemoryWarning = false;
```
fractalwrench, line 20:
```suggestion
        private bool _hasReceivedLowMemoryWarning = false;
```

src/BugsnagUnity/Native/Fallback/NativeClient.cs
```diff
@@ -12,18 +13,42 @@ class NativeClient : INativeClient
...
 
         public NativeClient(IConfiguration configuration)
         {
             Configuration = configuration;
             Breadcrumbs = new Breadcrumbs(configuration);
             Delivery = new Delivery();
+            Application.lowMemory += () => { _hasRecivedLowMemoryWarning = true; };
```
fractalwrench, line 27: Is there any way of detecting when the app exits the lowMemory state, or is this not possible with the available APIs?

fractalwrench (approved):
LGTM pending CI & resolution of comments

src/BugsnagUnity/Native/Fallback/NativeClient.cs
rich-bugsnag: thanks 👍

features/desktop/unhandled_errors.feature
rich-bugsnag: I wrote this following the example of other metadata tests. @twometresteve are there better ways i could be asserting here?

twometresteve: We do have an existing step that the Android (and probably others) use:
```    
And the event "device.time" is a timestamp
```

This Cucumber step is defined in MazeRunner and I can show you the docs that we have for them (they're not perfect, but could well help).

For Fractal's other suggestions, I'm not sure that we already have steps defined - but we can easily add them.  Let's talk IRL about that!

src/BugsnagUnity/Native/Fallback/NativeClient.cs
rich-bugsnag: Not possible unfortunatly

features/desktop/unhandled_errors.feature
rich-bugsnag: Added timestamp and non 0 int checks

PR #392, opened 2021-09-22
fix(device): make device id a notifier generated and stored value

rich-bugsnag:
## Goal

While rereading the notifier spec i noticed that i had implemented device.id incorectly.

## Changeset

- Made device.id a notifier generated and stored value for fallback

### DIFF
--- a/src/BugsnagUnity/Payload/Device.cs
+++ b/src/BugsnagUnity/Payload/Device.cs
@@ -12,6 +12,9 @@ namespace BugsnagUnity.Payload
     /// </summary>
     public class Device : Dictionary<string, object>, IFilterable
     {
+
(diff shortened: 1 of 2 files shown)


src/BugsnagUnity/Payload/Device.cs
```diff
@@ -33,7 +36,7 @@ internal Device(string hostname)
             this.AddToPayload("time", DateTime.UtcNow);
             AddBatteryLevel();
             this.AddToPayload("charging", SystemInfo.batteryStatus.Equals(BatteryStatus.Charging));
-            this.AddToPayload("id", SystemInfo.deviceUniqueIdentifier);
```
kattrali, line 36: > Will return SystemInfo.unsupportedIdentifier on platforms which don't support this property.

Ah TIL

kattrali (approved):

PR #396, opened 2021-09-24
Release v5.4.0

rich-bugsnag:
Release v5.4.0


tomlongridge (approved):
Not worth changing now, but in future (and for the GH release content) can we put the Unity enhancements before the changes from Cocoa & Android? I think they should get better prominence.

PR #421, opened 2021-11-01
Allow full build to be requested by the environment

twometresteve:
## Goal

Allow full build to be requested by the environment, for the purposes of schedules.


rich-bugsnag (approved):
thanks

PR #423, opened 2021-11-02
Bump minimum Unity version to 2018

twometresteve:
## Goal

Bump the minimum support version of Unity to 2018.

## Design

A change that comes with bumping the minimum supported version of Unity is that the notifier plugin will now be shipped built with Unity 2018.

## Changeset

- Notifier now built with Unity 2018 rather than 2017
- Running of tests on Unity 2017 removed from CI
- Mechanism for skipping flaky tests on Unity 2017 removed
- General updates to docs and helper scripts

## Testing

Covered by a full CI run.


.buildkite/pipeline.full.yml
```diff
@@ -133,7 +97,7 @@ steps:
     commands:
       - scripts/ci-run-macos-tests.sh
 
-  - label: Run MacOS e2e tests for Unity 2021
+  - label: Run macOS e2e tests for Unity 2021
```
fractalwrench, line 100: Will changing this label mean that the Github branch protection checks need updating?

docker-compose.yml
```diff
@@ -7,7 +7,6 @@ services:
       DEBUG:
       BUILDKITE:
       BUILDKITE_PIPELINE_NAME:
-      UNITY_VERSION:
```
fractalwrench, line 10: This still seems to be used elsewhere in the pipeline yaml: https://github.com/bugsnag/bugsnag-unity/pull/423/files#diff-a4cd708cc6050e0ba4f2289e6d602a54f79cba40363cbf890d8ff9ceade4c373R564

Is it definitely not required anymore?

fractalwrench (approved):
LGTM pending resolution of comments

.buildkite/pipeline.full.yml
twometresteve: I think we're ok as the status checks all seem to be lower-cased anyway.

docker-compose.yml
twometresteve: Good spot - the majority of uses in pipeline steps are in a non-Docker context (Mac and Windows build servers).  There was one instance where it was being passed to a Maze Runner container, but it was never required and presumably just a copy-paste error from early on (so I have removed it).

PR #427, opened 2021-11-05
Bump bugsnag-android to v5.15.0

lemnik:
## Goal
Bumps the bugsnag-android dependency to v5.15.0.


rich-bugsnag (changes requested):
Thanks @lemnik , plz could you add in the changelog entry?

rich-bugsnag (approved):
thanks!

PR #428, opened 2021-11-07
task(package): only build one 64 bit capable package

rich-bugsnag:
## Goal

With the launch of V6, we will no longer officially support unity 2017, which means we no longer need the extra 32 bit package for Android builds running in < unity 2017.4

## Changeset

- Made the rake file only output the 64 bit version
- Refactored the fixture build scripts to not expect the 64 bit named package
- Removed any artifact mentions of the old package in the CI pipeline 

## Testing

Manually tested in builds and passing CI Tests should indicate everything working well

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -328,16 +328,16 @@ namespace :plugin do
 
   desc "Generate release artifacts"
   task export: ["plugin:build:clean"] do
-    Rake::Task["plugin:build:all"].invoke
-    export_package("Bugsnag.unitypackage")
+    #Rake::Task["plugin:build:all"].invoke
+    #export_package("Bugsnag.unitypackage")
     Rake::Task["plugin:build:all_android64"].invoke
-    export_package("Bugsnag-with-android-64bit.unitypackage")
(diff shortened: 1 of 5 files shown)


TESTING.md
```diff
@@ -97,9 +97,8 @@ To build any test fixture, from the root of the repository, first build the noti
 ```
 rake plugin:export
 ```
-This will generate the following files:
+This will generate the following file:
```
twometresteve, line 100: See also line 20.

Rakefile
```diff
@@ -328,16 +328,16 @@ namespace :plugin do
 
   desc "Generate release artifacts"
   task export: ["plugin:build:clean"] do
-    Rake::Task["plugin:build:all"].invoke
-    export_package("Bugsnag.unitypackage")
+    #Rake::Task["plugin:build:all"].invoke
+    #export_package("Bugsnag.unitypackage")
```
twometresteve, line 332: Remove these lines.

twometresteve (approved):
Couple of minor comments, but happy for it to be merged without re-review.

PR #436, opened 2021-11-23
task(cocoa): review all code in cocoa bridge class

rich-bugsnag:
## Goal

Do a full review of the file BugsnagUnity.mm


src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+#import "Bugsnag+Private.h"
+#import "BugsnagClient+Private.h"
+#import "BSG_KSSystemInfo.h"
+#import "BSG_KSMach.h"
+#import "BSG_KSCrash.h"
+#import "BSG_RFC3339DateTool.h"
+#import "BugsnagBreadcrumb+Private.h"
+#import "BugsnagSession+Private.h"
+
+extern "C" {
```
nickdowell, line 16: If we make this a `.m` instead of `.mm` file, we can remove the whole `extern "C"` section (we'll need to keep the definition of `struct bugsnag_user` though)

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+        NSData *data = [NSData dataWithBytesNoCopy:(void *)jsonString length:strlen(jsonString) freeWhenDone:NO];
+        id metadata = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
+        if ([metadata isKindOfClass:[NSDictionary class]]) {
+            return metadata;
+        }
+    } @catch (NSException *exception) {
+        NSLog(@"%@", exception);
+        return nil;
+    }
+}
```
nickdowell, line 203: This function is missing a return value in the case that no exception is thrown but metadata is not a dictionary.

```suggestion
    } @catch (NSException *exception) {
        NSLog(@"%@", exception);
    }
    return nil;
}
```

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+    NSDictionary* sectionDictionary = ((__bridge BugsnagBreadcrumb *)breadcrumb).metadata;
+    return getMetadataJson(sectionDictionary);
+}
+
+void bugsnag_setBreadcrumbMetadata(const void *breadcrumb, const char *jsonString) {
+  if (jsonString == NULL)
+  {
+    return;
+  }
+  ((__bridge BugsnagBreadcrumb *)breadcrumb).metadata = getDictionaryFromMetadataJson(jsonString);
```
nickdowell, line 330: Shouldn't this allow setting metadata to NULL?

```suggestion
  ((__bridge BugsnagBreadcrumb *)breadcrumb).metadata = jsonString  ? getDictionaryFromMetadataJson(jsonString) : nil;
```

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+
+void bugsnag_markLaunchCompleted() {
+  [Bugsnag markLaunchCompleted];
+}
+
+void *bugsnag_createConfiguration(char *apiKey) {
+  NSString *ns_apiKey = [NSString stringWithUTF8String: apiKey];
+  BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:ns_apiKey];
+  config.apiKey = ns_apiKey;
+  return (void*)CFBridgingRetain(config);
```
nickdowell, line 511:
```suggestion
  return (void *)CFBridgingRetain([[BugsnagConfiguration alloc] initWithApiKey:@(apiKey)]);
```

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+
+void bugsnag_setNotifyReleaseStages(const void *configuration, const char *releaseStages[], int releaseStagesCount){
+  NSMutableSet *ns_releaseStages = [NSMutableSet new];
+  for (int i = 0; i < releaseStagesCount; i++) {
+    const char *releaseStage = releaseStages[i];
+    if (releaseStage != nil) {
+      NSString *ns_releaseStage = [NSString stringWithUTF8String: releaseStage];
+      [ns_releaseStages addObject: ns_releaseStage];
+    }
+  }
```
nickdowell, line 527: This conversion from array of C strings to an NSSet is repeated a few times (also in `bugsnag_setDiscardClasses` and `bugsnag_setRedactedKeys`) and would be good to extract to a reusable function.

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+  if (notifierVersion) {
+    ((__bridge BugsnagConfiguration *)configuration).notifier =
+      [[BugsnagNotifier alloc] initWithName:@"Unity Bugsnag Notifier"
+                                    version:@(notifierVersion)
+                                        url:@"https://github.com/bugsnag/bugsnag-unity"
+                               dependencies:@[[[BugsnagNotifier alloc] init]]];
+  }
+  [Bugsnag startWithConfiguration: (__bridge BugsnagConfiguration *)configuration];
+  // Memory introspection is unused in a C/C++ context
+  [BSG_KSCrash sharedInstance].introspectMemory = NO;
```
nickdowell, line 764: This is not necessary - it's set to NO by default now.

```suggestion
  // Memory introspection is unused in a C/C++ context
  [BSG_KSCrash sharedInstance].introspectMemory = NO;
```

src/BugsnagUnityForReview.mm
```diff
@@ -0,0 +1,876 @@
...
+void bugsnag_setUser(char *userId, char *userName, char *userEmail) {
+    [Bugsnag setUser:userId == NULL ? nil : [NSString stringWithUTF8String:userId]
+            withEmail:userEmail == NULL ? nil : [NSString stringWithUTF8String:userEmail]
+             andName:userName == NULL ? nil : [NSString stringWithUTF8String:userName]];
+}
+
+@interface Bugsnag ()
++ (BugsnagNotifier *)notifier;
+@end
+
```
nickdowell, line 865: This should be deleted because it declares a method that doesn't exist.

```suggestion
```

nickdowell (commented):
Didn't spot any serious causes for concern.

`bugsnag_retrieveAppData()` and `bugsnag_retrieveDeviceData()` do make assumptions about the type of objects retrieved from NSDictionaries, which could potentially lead to unhandled exceptions, but the risk is low due to `BSG_KSSystemInfo` changing very infrequently. Any issues should get identified immediately through automated testing, since any call to these functions would trigger the issues.

src/BugsnagUnityForReview.mm
rich-bugsnag: Done 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: Done 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: Done 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: Removed 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: Removed 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: Done 👍

src/BugsnagUnityForReview.mm
rich-bugsnag: done and seems to work great 👍

nickdowell (approved):

PR #444, opened 2021-12-02
task(callbacks): add android session callbacks Part 1

rich-bugsnag:
## Goal

This is Part 1 of the larger Unity/Android callbacks task. 

I'm breakign it down into small chunks because it's going to be a lot of changes. 

In this part, i setup the connection between Android and Unity and created a basic (incomplete) NativeSession wrapper class and NativeUser wrapper class

## Changeset

- Hooked into the Android layer session callbacks in the NativeInterface.cs class
- Created the Wrapper classes NativeSession and NativeUser
- Fixed an oversight in how Cocoa sessions and events called SetUser

## Testing

Manually tested, CI will be added at the end of the task

### DIFF
--- a/src/BugsnagUnity.m
+++ b/src/BugsnagUnity.m
@@ -80,8 +80,20 @@ BugsnagUser * bugsnag_getUserFromSession(const void *session){
     return ((__bridge BugsnagSession *)session).user;
 }
 
+void bugsnag_setUserFromSession(const void *session, char *userId, char *userName, char *userEmail){
+    [((__bridge BugsnagSession *)session) setUser:userId == NULL ? nil : [NSString stringWithUTF8String:userId]
+            withEmail:userEmail == NULL ? nil : [NSString stringWithUTF8String:userEmail]
+             andName:userName == NULL ? nil : [NSString stringWithUTF8String:userName]];
+}
+
 BugsnagUser * bugsnag_getUserFromEvent(const void *event){
     return ((__bridge BugsnagEvent *)event).user;
+}
+
+void bugsnag_setUserFromEvent(const void *event, char *userId, char *userName, char *userEmail){
+    [((__bridge BugsnagEvent *)event) setUser:userId == NULL ? nil : [NSString stringWithUTF8String:userId]
+            withEmail:userEmail == NULL ? nil : [NSString stringWithUTF8String:userEmail]
+             andName:userName == NULL ? nil : [NSString stringWithUTF8String:userName]];
 }
 
 void bugsnag_getThreadsFromEvent(const void *event,const void *instance, void (*callback)(const void *instance, void *threads[], int threads_size)) {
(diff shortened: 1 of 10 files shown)


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -49,6 +49,29 @@ class NativeInterface
...
+                var wrapper = new NativeSession(session);
+                foreach (var callback in _config.GetOnSessionCallbacks())
+                {
+                    if (!callback.Invoke(wrapper))
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            }
```
kattrali, line 72: Implementing a Java interface with a C# class is pure magic

src/BugsnagUnity.m
```diff
@@ -80,10 +80,22 @@ void bugsnag_setEventMetadata(const void *event, const char *tab, const char *me
...
 BugsnagUser * bugsnag_getUserFromEvent(const void *event){
     return ((__bridge BugsnagEvent *)event).user;
 }
 
+void bugsnag_setUserFromEvent(const void *event, char *userId, char *userName, char *userEmail){
+    [((__bridge BugsnagEvent *)event) setUser:userId == NULL ? nil : [NSString stringWithUTF8String:userId]
+            withEmail:userEmail == NULL ? nil : [NSString stringWithUTF8String:userEmail]
+             andName:userName == NULL ? nil : [NSString stringWithUTF8String:userName]];
+}
+
```
kattrali, line 98: I'm going to guess that these might need some tests? Its weird that they could be completely absent without error.

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -49,6 +49,29 @@ class NativeInterface
...
+            public OnSessionCallback(Configuration config) : base("com.bugsnag.android.OnSessionCallback")
+            {
+                _config = config;
+            }
+            public bool onSession(AndroidJavaObject session)
+            {
+                var wrapper = new NativeSession(session);
+                foreach (var callback in _config.GetOnSessionCallbacks())
+                {
+                    if (!callback.Invoke(wrapper))
```
fractalwrench, line 66: If the callback throws an exception the requirement is to catch it and continue if possible. I think there are also some requirements around logging information - the spec should have more info.

fractalwrench (approved):
LGTM pending resolution of comment

src/BugsnagUnity/Native/Android/NativeInterface.cs
rich-bugsnag: 🧙

src/BugsnagUnity/Native/Android/NativeInterface.cs
rich-bugsnag: Added a handler, could not find any mention of logging in the spec

src/BugsnagUnity.m
kattrali: _Resolving via discussion of the tests in subsequent changeset_

kattrali (approved):

PR #445, opened 2021-12-02
docs(upgrade guide): improve readability

tomlongridge:
## Goal

I've reworked the Upgrade Guide content so that the Key Points listed match 1:1 with the sections below.

Also reformatted some of the sections to make it easier to read, with a summary of the deprecations at the bottom for clarity.


rich-bugsnag (approved):
Thanks!

PR #446, opened 2021-12-03
Allow retry on build steps for lost agents and timeouts

twometresteve:
## Goal

Allow retry on build (but not test) steps for lost agents and timeouts.

## Design

Allowing retries on test steps as well feels risky, as we might might miss a genuine, intermittent, issue.

## Changeset

Only an update to the CI pipeline to allow retries.

## Testing

Covered by CI.


rich-bugsnag (approved):
Ahhhh yeahhhh!

PR #447, opened 2021-12-06
task(callbacks): add android session callbacks Part 2

rich-bugsnag:
## Goal

This is the 2nd and final part of wrapping Android session callbacks

## Changeset

- Added NativeApp and NativeDevice wrappers
- Added implementations for all NativeSession properties

## Testing

Added full CI tests in features/android/android_callbacks.feature

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeApp.cs
+++ b/src/BugsnagUnity/Native/Android/NativeApp.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using BugsnagUnity.Payload;
+using UnityEngine;
+
+namespace BugsnagUnity
+{
+    public class NativeApp : NativePayloadClassWrapper, IApp
+    {
+        public NativeApp(AndroidJavaObject nativePointer) : base (nativePointer){}
+
+        public string BinaryArch { get => GetNativeString("getBinaryArch"); set => SetNativeString("setBinaryArch", value); }
+        public string BuildUuid { get => GetNativeString("getBuildUuid"); set => SetNativeString("setBuildUuid", value); }
+        public string CodeBundleId { get => GetNativeString("getCodeBundleId"); set => SetNativeString("setCodeBundleId", value); }
+        public string Id { get => GetNativeString("getId"); set => SetNativeString("setId", value); }
+        public string ReleaseStage { get => GetNativeString("getReleaseStage"); set => SetNativeString("setReleaseStage", value); }
+        public string Type { get => GetNativeString("getType"); set => SetNativeString("setType", value); }
+        public string Version { get => GetNativeString("getVersion"); set => SetNativeString("setVersion", value); }
+        public int? VersionCode { get => GetNativeInt("getVersionCode"); set => SetNativeInt("setVersionCode", value); }
+
+        public string BundleVersion { get; set; }
+        public string DsymUuid { get; set; }
+
+
+    }
+}
--- a/src/BugsnagUnity/Native/Android/NativeApp.cs
+++ b/src/BugsnagUnity/Native/Android/NativeApp.cs
@@ -4,9 +4,9 @@ using UnityEngine;
 
 namespace BugsnagUnity
 {
-    public class NativeApp : NativePayloadClassWrapper, IApp
+    internal class NativeApp : NativePayloadClassWrapper, IApp
     {
-        public NativeApp(AndroidJavaObject nativePointer) : base (nativePointer){}
+        internal NativeApp(AndroidJavaObject nativePointer) : base (nativePointer){}
 
         public string BinaryArch { get => GetNativeString("getBinaryArch"); set => SetNativeString("setBinaryArch", value); }
         public string BuildUuid { get => GetNativeString("getBuildUuid"); set => SetNativeString("setBuildUuid", value); }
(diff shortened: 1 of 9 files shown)


src/BugsnagUnity/Native/Android/NativeApp.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using BugsnagUnity.Payload;
+using UnityEngine;
+
+namespace BugsnagUnity
+{
+    public class NativeApp : NativePayloadClassWrapper, IApp
```
fractalwrench, line 7: Will making this `public` expose it to users as part of our API?

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -853,7 +853,7 @@ private AndroidJavaObject BuildJavaStringDisposable(string input)
             }
         }
 
-        private bool IsUnity2019OrNewer()
+        internal static bool IsUnity2019OrNewer()
```
fractalwrench, line 856: Do these definitely need to be static? If it's unavoidable then we can go down this route, but it feels like it'd be preferable to keep things non-static (this will make it much easier to pass in a mock NativeInterface to other classes etc)

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
```diff
@@ -22,5 +23,183 @@ public void SetNativeString(string key, string value)
...
+                NativePointer.Call(key, null);
+            }
+            else
+            {
+                var javaBoolean = new AndroidJavaObject("java.lang.Boolean", (bool)value);
+                NativePointer.Call(key, javaBoolean);
+            }
+        }
+
+        public string[] GetNativeStringArray(string key)
```
fractalwrench, line 89: It might be worth returning `List<string>` here (or another method that does the equivalent). I'm wary that this might lead to a situation where a collection is editable on Android, but remains a static size in C#

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
```diff
@@ -22,5 +23,183 @@ public void SetNativeString(string key, string value)
...
+
+                    var iterator = keys.Call<AndroidJavaObject>("iterator");
+
+                    var dict = new Dictionary<string, object>();
+
+                    while (iterator.Call<bool>("hasNext"))
+                    {
+                        var next = iterator.Call<AndroidJavaObject>("next");
+                        var theKey = next.Call<string>("toString");
+                        var theValue = map.Call<AndroidJavaObject>("get",theKey);
```
fractalwrench, line 155: Would probably be worth adding a null check - some maps implementations on the JVM can hold null values

src/BugsnagUnity/Native/Android/NativeDevice.cs
```diff
@@ -0,0 +1,30 @@
...
+        public string UserAgent { get; set; }
+
+
+        public string Id { get => GetNativeString("getId"); set => SetNativeString("setId", value); }
+        public string Locale { get => GetNativeString("getLocale"); set => SetNativeString("setLocale", value); }
+        public string Manufacturer { get => GetNativeString("getManufacturer"); set => SetNativeString("setManufacturer", value); }
+        public string Model { get => GetNativeString("getModel"); set => SetNativeString("setModel", value); }
+        public string OsName { get => GetNativeString("getOsName"); set => SetNativeString("setOsName", value); }
+        public string OsVersion { get => GetNativeString("getOsVersion"); set => SetNativeString("setOsVersion", value); }
+        public int? TotalMemory { get => GetNativeInt("getTotalMemory"); set => SetNativeLong("setTotalMemory", value); }
```
kattrali, line 23: I don't think this is guaranteed to fit in an int, and you might get weird behavior when you try to get or set a large number.

kattrali (commented):
One small issue, otherwise looks good to me!

src/BugsnagUnity/Native/Android/NativeApp.cs
rich-bugsnag: Changed to internal

src/BugsnagUnity/Native/Android/NativeInterface.cs
rich-bugsnag: I think this is ok for now as i don't want to duplicate code

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
rich-bugsnag: Another method for mutable collections is coming in the next PR

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
rich-bugsnag: Done

src/BugsnagUnity/Native/Android/NativeDevice.cs
rich-bugsnag: This will be addressed in the next PR

kattrali (approved):

fractalwrench (approved):

PR #448, opened 2021-12-08
Update bugsnag-android to v5.17.0

fractalwrench:
## Goal

Updates bugsnag-android dependency to v5.17.0.


rich-bugsnag (approved):
Thanks!

PR #450, opened 2021-12-10
Increase WebGL memory size for example and test fixture

twometresteve:
Increase WebGL memory size in the example and e2e test fixture, following some recent out of memory errors in our e2e tests with Unity 2021.


rich-bugsnag (approved):
Thanks!

PR #452, opened 2021-12-10
refac(callbacks): replace named delegates with Func

rich-bugsnag:
## Goal

replace named delegates with Func

## Changeset

- Changed all error type callbacks to Func<IEvent, bool>
- Changed all session type callbacks to Func<ISession, bool>

### DIFF
--- a/src/BugsnagUnity/IClient.cs
+++ b/src/BugsnagUnity/IClient.cs
@@ -47,7 +47,7 @@ namespace BugsnagUnity
 
         void AddOnError(Func<IEvent, bool> callback);
 
-        void RemoveOnError(Func<IEvent, bool> bugsnagCallback);
+        void RemoveOnError(Func<IEvent, bool> callback);
 
         void AddOnSession(Func<ISession, bool> callback);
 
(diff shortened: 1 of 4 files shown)


src/BugsnagUnity/IClient.cs
```diff
@@ -50,13 +45,13 @@ internal interface IClient : IMetadataEditor
 
         void MarkLaunchCompleted();
 
-        void AddOnError(OnErrorCallback bugsnagCallback);
+        void AddOnError(Func<IEvent, bool> bugsnagCallback);
```
gareththackeray, line 48:
```suggestion
        void AddOnError(Func<IEvent, bool> callback);
```

src/BugsnagUnity/IClient.cs
```diff
@@ -50,13 +45,13 @@ internal interface IClient : IMetadataEditor
...
 
-        void RemoveOnError(OnErrorCallback bugsnagCallback);
+        void RemoveOnError(Func<IEvent, bool> bugsnagCallback);
```
gareththackeray, line 50:
```suggestion
        void RemoveOnError(Func<IEvent, bool> callback);
```

gareththackeray (approved):

PR #455, opened 2021-12-13
refac(dateTime): change to DateTimeOffset

rich-bugsnag:
## Goal

Change usage of DateTime to DateTimeOffset

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
+++ b/src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
@@ -251,7 +251,7 @@ namespace BugsnagUnity
             }
         }
 
-        public DateTime? GetNativeDateTime(string key)
+        public DateTimeOffset? GetNativeDateTime(string key)
         {
             var nativeDate = NativePointer.Call<AndroidJavaObject>(key);
 
@@ -267,7 +267,7 @@ namespace BugsnagUnity
             return date;
         }
 
-        public void SetNativeDateTime(string key, DateTime? dateTime)
+        public void SetNativeDateTime(string key, DateTimeOffset? dateTime)
         {
             if (dateTime == null)
             {
--- a/src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
+++ b/src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
@@ -262,7 +262,7 @@ namespace BugsnagUnity
 
             var timeStamp = nativeDate.Call<long>("getTime");
 
-            var date = (new DateTime(1970, 1, 1)).AddMilliseconds(timeStamp);
+            var date = new DateTimeOffset(1970, 1, 1,0,0,0, TimeSpan.Zero).AddMilliseconds(timeStamp);
 
             return date;
         }
@@ -275,7 +275,7 @@ namespace BugsnagUnity
             }
             else
             {
-                var mill = (dateTime - new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc)).Value.TotalMilliseconds;
+                var mill = (dateTime - new DateTimeOffset(1970, 1, 1, 0, 0, 0, TimeSpan.Zero)).Value.TotalMilliseconds;
                 var javaDate = new AndroidJavaObject("java.util.Date", (long)mill);
                 NativePointer.Call(key,javaDate);
             }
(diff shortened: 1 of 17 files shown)


gareththackeray:
By using `DateTimeOffset.Now` rather than `DateTimeOffset.UtcNow` we'll be changing what we send over the wire.  (Because it'll include the user's local offset rather than `z`.)  I don't think this is an issue since it's a major version but we should mention in the release notes.

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
```diff
@@ -267,7 +267,7 @@ public void SetNativeMetadataSection(string key, string section, Dictionary<stri
             return date;
```
gareththackeray, line 267: I think we should construct a `DateTimeOffset` on the line above rather than relying on an implicit cast.

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
```diff
@@ -267,7 +267,7 @@ public void SetNativeMetadataSection(string key, string section, Dictionary<stri
...
         }
 
-        public void SetNativeDateTime(string key, DateTime? dateTime)
+        public void SetNativeDateTime(string key, DateTimeOffset? dateTime)
         {
             if (dateTime == null)
             {
```
gareththackeray, line 273: Likewise below, let's construct a `DateTimeOffset` and do the arithmetic using that.

src/BugsnagUnity/Native/Cocoa/NativeDeviceWithState.cs
```diff
@@ -22,6 +22,6 @@ public NativeDeviceWithState(IntPtr nativeDevice) : base(nativeDevice)
 
         public string Orientation { get => NativeWrapper.GetNativeString(ORIENTATION_KEY); set => NativeWrapper.SetNativeString(ORIENTATION_KEY,value); }
 
-        public DateTime? Time { get => DateTime.Parse(NativeWrapper.GetNativeString(TIME_KEY)); set => NativeWrapper.SetNativeString(TIME_KEY, value?.ToLongTimeString()); }
+        public DateTimeOffset? Time { get => DateTimeOffset.Parse(NativeWrapper.GetNativeString(TIME_KEY)); set => NativeWrapper.SetNativeString(TIME_KEY, value?.ToString()); }
```
gareththackeray, line 25: Is `ToString()` here guaranteed to give us the same format regardless of the user's local culture?  I think using ISO format (`"o"`) would be safer?

src/BugsnagUnity/Native/Cocoa/NativePayloadClassWrapper.cs
```diff
@@ -83,13 +83,13 @@ internal void SetNativeLong(string key, long? value)
             }
             else
             {
-                DateTime dateTime = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
+                DateTimeOffset dateTime = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
```
gareththackeray, line 86: Let's construct a `DateTimeOffset` explicitly.

src/BugsnagUnity/Native/Cocoa/NativePayloadClassWrapper.cs
```diff
@@ -83,13 +83,13 @@ internal void SetNativeLong(string key, long? value)
...
                 dateTime = dateTime.AddSeconds(timeStamp);
                 return dateTime;
             }
         }
 
-        internal void SetNativeDate(DateTime? startedAt,string key)
+        internal void SetNativeDate(DateTimeOffset? startedAt,string key)
         {
             if (startedAt == null)
             {
```
gareththackeray, line 95: Same below.

gareththackeray (commented):
We should remove all the internal usages of `DateTime` as well as it's just less to think about.

src/BugsnagUnity/Native/Cocoa/NativeDeviceWithState.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/Native/Android/NativePayloadClassWrapper.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/Native/Cocoa/NativePayloadClassWrapper.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/Native/Cocoa/NativePayloadClassWrapper.cs
rich-bugsnag: Done 👍

gareththackeray (approved):

PR #458, opened 2021-12-14
fix(breadcrumbs): fix breadcrumbs not being editable in callbacks

rich-bugsnag:
## Goal

Fallback Breadcrumbs were not up to spec and this was causing confusion when trying to edit crumbs in an event callback

## Changeset

- Refactored Breadcrumb.cs to be up to spec
- Updated the automatic recording of log and error breadcrumbs to be up to spec

## Testing

Improved existing desktop tests

### DIFF
--- a/src/BugsnagUnity/Payload/Breadcrumb.cs
+++ b/src/BugsnagUnity/Payload/Breadcrumb.cs
@@ -9,39 +9,37 @@ namespace BugsnagUnity.Payload
     /// </summary>
     public class Breadcrumb : PayloadContainer, IBreadcrumb
     {
-        private const string UNDEFINED_NAME = "Breadcrumb";
-        private const string NAME_KEY = "name";
-        private const string MESSAGE_KEY = "message";
+        private const string MESSAGE_KEY = "name";
         private const string TIMESTAMP_KEY = "timestamp";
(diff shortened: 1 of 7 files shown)


src/BugsnagUnity/Payload/Breadcrumb.cs
```diff
@@ -9,39 +9,37 @@ namespace BugsnagUnity.Payload
     /// </summary>
     public class Breadcrumb : PayloadContainer, IBreadcrumb
     {
-        private const string UNDEFINED_NAME = "Breadcrumb";
-        private const string NAME_KEY = "name";
-        private const string MESSAGE_KEY = "message";
+        private const string MESSAGE_KEY = "name";
```
gareththackeray, line 12: This seems surprising on the face of it.  Why `name`?

gareththackeray (commented):
Not completely following this.   Does it allow insertion / deletion? Should it?

src/BugsnagUnity/Payload/Breadcrumb.cs
rich-bugsnag: added a comment to explain

gareththackeray (approved):

PR #459, opened 2021-12-15
task(init): add warning if already initialised

rich-bugsnag:
## Goal

add warning if already initialised

## Changeset

- Added warning to the Bugsnag.Start method

## Testing

Manually tested

### DIFF
--- a/src/BugsnagUnity/Bugsnag.cs
+++ b/src/BugsnagUnity/Bugsnag.cs
@@ -7,7 +7,7 @@ namespace BugsnagUnity
     public static class Bugsnag
     {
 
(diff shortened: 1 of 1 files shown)


src/BugsnagUnity/Bugsnag.cs
```diff
@@ -6,6 +6,9 @@ namespace BugsnagUnity
 {
     public static class Bugsnag
     {
+
+        private const string INIT_WARNING = "Bugsnag is already running and this call to Start() will be ignored. Please check whether Bugsnag is set to start automatically via the settings dialog.";
```
gareththackeray, line 10:
```suggestion
        private const string INIT_WARNING = "Bugsnag is already running and this call to Start() will be ignored. If this was unexpected please check whether Bugsnag is set to start automatically via the settings dialog.";
```

gareththackeray (approved):

PR #469, opened 2021-12-17
task(settings): change default values, renamed threadsend enum

rich-bugsnag:
## Goal

- Show default endpoints in settings window
- rename threadsend enum to pascal case
- remove default production release stage
- remove legacy standard assets folder from the example project 
- change default max breadcrumbs to 50


tomlongridge (approved):
LGTM - pending CI

PR #471, opened 2021-12-20
fix(config): launch duration 0 fix

rich-bugsnag:
## Goal

fix issue where launch duration 0 was not resulting in IsLaunching true even though MarkLaunchComp[lete was not called

### DIFF
--- a/src/BugsnagUnity/Native/Fallback/NativeClient.cs
+++ b/src/BugsnagUnity/Native/Fallback/NativeClient.cs
@@ -42,7 +42,7 @@ namespace BugsnagUnity
             bool isLaunching;
             if (Configuration.LaunchDurationMillis == 0)
             {
-                isLaunching = _launchMarkedAsCompleted;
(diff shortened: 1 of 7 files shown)


src/BugsnagUnity/Native/Fallback/NativeClient.cs
```diff
@@ -42,7 +42,7 @@ private void AddIsLaunching(AppWithState app)
             bool isLaunching;
             if (Configuration.LaunchDurationMillis == 0)
             {
-                isLaunching = _launchMarkedAsCompleted;
+                isLaunching = !_launchMarkedAsCompleted;
```
gareththackeray, line 45: This one feels like something an automated test should have found?

rich-bugsnag: added tests for both setting 0 and setting 0 and calling MarkLaunchComplete 👍

gareththackeray (approved):

PR #472, opened 2021-12-21
fix(metadata): fix ClearMetadata

rich-bugsnag:
## Goal

ClearMetadata was not working as intended and test covarage needed improving

## Changeset

- Added the following methods to the INativeClient interface: ClearNativeMetadata, SetNativeMetadata, GetNativeMetadata to simplify the code and make it easier to understand the relationship between Native and c# metadata.
- Added new tests for ClearMetadata

## Testing

- Added new tests for ClearMetadata

### DIFF
--- a/features/android/android_metadata.feature
+++ b/features/android/android_metadata.feature
@@ -1,0 +1,21 @@
+Feature: Android smoke tests for Metadata
+
+    Background:
+        Given I wait for the mobile game to start
+
+    Scenario: Clear Metadata
+        When I run the "Clear Metadata" mobile scenario
+        Then I wait to receive an error
+       
+        # MetaData
+        And the event "metaData.test.test1" equals "test1"
+        And the event "metaData.test.test2" is null
+        And the event "metaData.test3.test3" equals "test3"
+        And the event "metaData.test4" is null
+
+
+
+
+
+
+
--- a/features/ios/ios_metadata.feature
+++ b/features/ios/ios_metadata.feature
@@ -1,0 +1,21 @@
+Feature: Android smoke tests for Metadata
+
+    Background:
+        Given I wait for the mobile game to start
+
+    Scenario: Clear Metadata
+        When I run the "Clear Metadata" mobile scenario
+        Then I wait to receive an error
+       
+        # MetaData
+        And the event "metaData.test.test1" equals "test1"
+        And the event "metaData.test.test2" is null
+        And the event "metaData.test3.test3" equals "test3"
+        And the event "metaData.test4" is null
+
+
+
+
+
+
+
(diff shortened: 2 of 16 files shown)


features/android/android_metadata.feature
```diff
@@ -0,0 +1,21 @@
...
+        And the event "metaData.test.test1" equals "test1"
+        And the event "metaData.test.test2" is null
+        And the event "metaData.test3.test3" equals "test3"
+        And the event "metaData.test4" is null
+
+
+
+
+
+
```
fractalwrench, line 20:
```suggestion
        And the event "metaData.test4" is null
```

features/ios/ios_metadata.feature
```diff
@@ -0,0 +1,22 @@
...
+        And the event "metaData.test.test2" is null
+        And the event "metaData.test3.test3" equals "test3"
+        And the event "metaData.test4" is null
+
+
+
+
+
+
+
```
fractalwrench, line 22:
```suggestion

```

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -544,6 +553,8 @@ public void AddToTab(string tab, string key, string value)
                 new object[] { tab, key, value });
         }
 
+       
+
```
fractalwrench, line 557:
```suggestion
```

src/BugsnagUnity/Native/Android/NativeClient.cs
```diff
@@ -132,6 +112,28 @@ public LastRunInfo GetLastRunInfo()
...
+        {
+            NativeInterface.ClearMetadata(section,key);
+        }
+
+        public IDictionary<string, object> GetNativeMetadata()
+        {
+            return NativeInterface.GetMetadata();
+        }
+
+        public void SetNativeMetadata(string section, IDictionary<string, object> data)
```
fractalwrench, line 130: I'm not sure if this function makes sense - I would assume that this clears any existing metadata, then adds all the key-value pairs in `data`. Right now the implementation just adds metadata, so it feels like the values could be wrong if someone forgot to call `ClearNativeMetadata` first?

src/BugsnagUnity/Payload/Metadata.cs
```diff
@@ -32,61 +33,45 @@ public void AddMetadata(string section, IDictionary<string, object> metadataSect
             }
             if (SectionExists(section))
             {
-
                 var existingSection = (IDictionary<string, object>)Get(section);
 
-                foreach (var entry in metadataSection)
+                foreach (var key in metadataSection.Keys.ToList())
```
fractalwrench, line 38: Is there an advantage to using the keys directly rather than an entryset?

rich-bugsnag: It allows for the set to be edited while iterating

src/BugsnagUnity/Native/Android/NativeClient.cs
rich-bugsnag: The behaviour is correct, but the name was missleading, i have renamed it 👍

fractalwrench (approved):
LGTM

PR #473, opened 2021-12-22
Use Maze.check abstraction for making test assertions

twometresteve:
## Goal

Use the `Maze.check` abstraction just added to Maze Runner rather than `assert` functions from the global Ruby namespace (which can be problematic and makes it harder to alter the behaviour of checks if we so wish).

## Changeset

Also corrected a number of Rubocop style violations as part of this change.

## Testing

Covered by a basic CI run.


rich-bugsnag (approved):
Thanks!

PR #474, opened 2021-12-22
fix(config): set BugsnagAppHangThresholdFatalOnly

rich-bugsnag:
## Goal

set BugsnagAppHangThresholdFatalOnly when AppHangThresholdMillis is set to 0


fractalwrench (approved):
LGTM

PR #476, opened 2021-12-23
task(settings): add tooltips to the android and apple only options

rich-bugsnag:
## Goal

Add tooltips to the android an apple only setting lablels

Also update the gitignore in the example project


fractalwrench (approved):
LGTM

PR #478, opened 2022-01-04
Comment to explain what unit is when calling "sleep()"

gareththackeray:
The unit of `sleep()` is s, not ms.


rich-bugsnag (approved):
Thanks!

PR #481, opened 2022-01-06
Correct scene file paths

twometresteve:
## Goal

Correct the scene file paths so that the entries in the custom Build menu for the example project work.

## Testing

Tested locally.


rich-bugsnag (approved):
Thanks

PR #482, opened 2022-01-06
Mention change to Context + couple of typos

gareththackeray:


UPGRADING.md
```diff
@@ -60,6 +61,9 @@ An `Bugsnag.AddOnSendError()` allows you to add a C# function that will be execu
 
 See our [online docs](https://docs.bugsnag.com/platforms/unity/customizing-error-reports/) for further information.
 
+### Default context
+The `Context` of an event is no longer set by default to the current scene name. If you want to continue having `Context` set to the scene name you should set `Bugsnag.Context = <scene name>;` when it changes.
```
gareththackeray, line 65: @rich-bugsnag is this the right way to do it?  Can we be more precise?
@tomlongridge I haven't mentioned the impact it could have if you're grouping errors by class for particular contexts.  It's a very little used feature and I feel mentioning it might be more confusing.  What do you think?

gareththackeray:
```suggestion
The `Context` of an event is no longer set by default to the current scene name. This will change how errors are displayed on your dashboard and may affect the grouping of errors if you are using the "Group by error context" project setting. If you want to continue having `Context` set to the scene name you should set `Bugsnag.Context = <scene name>;` when it changes.
```

rich-bugsnag: Maybe change it to "no longer set by default to the last loaded scene name" as that's the behaviour. Apart from that i think it's good 👍

gareththackeray:
```suggestion
The `Context` of an event is no longer set by default to the last loaded scene name. This will change how errors are displayed on your dashboard and may affect the grouping of errors if you are using the "Group by error context" project setting. If you want to continue having `Context` set to the last loaded scene name you should set `Bugsnag.Context = <scene name>;` upon loading scenes.
```

tomlongridge: It's probably worth appending that this is now captured in metadata as `app.lastLoadedUnityScene`.

gareththackeray:
```suggestion
The `Context` of an event is no longer set by default to the last loaded scene name. This will change how errors are displayed on your dashboard and may affect the grouping of errors if you are using the "Group by error context" project setting. If you want to continue having `Context` set to the last loaded scene name you should set `Bugsnag.Context = <scene name>;` on scene loads.

The last loaded scene name is now captured in metadata as `app.lastLoadedUnityScene`. 
```

rich-bugsnag (approved):
Thanks 👍

tomlongridge (approved):

PR #492, opened 2022-01-13
Add the active unity scene to the event metadata

rich-bugsnag:
## Goal
Add active unity scene metadata

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -351,6 +351,8 @@ namespace BugsnagUnity
 
             AutomaticDataCollector.AddStatefulDeviceData(eventMetadata);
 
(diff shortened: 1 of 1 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -351,6 +351,8 @@ private void NotifyOnMainThread(Error[] exceptions, HandledState handledState, F
 
             AutomaticDataCollector.AddStatefulDeviceData(eventMetadata);
 
+            eventMetadata.AddMetadata("app", "activeUnityScene", SceneManager.GetActiveScene().name);
```
gareththackeray, line 354: `GetActiveScene()` can't return null?

gareththackeray (approved):

PR #494, opened 2022-01-19
Merge V6 into next for V6 release

rich-bugsnag:
## Goal

Merge V6 into next for V6 release


tomlongridge (approved):
There's currently no changelog entry for this, but given CI has run, I'm happy if you want to do that on your release PR given that it will largely be pointing at the Upgrade Guide.

PR #498, opened 2022-01-21
task(CI): enable 64 bit IL2CPP android tests

rich-bugsnag:
## Goal

Until now we have been building the android fixtures with mono, outputing 32bit APKs.

Mono and 32bit are now out of date as 64 bit builds are a requirement for submitting to the google play store.

This change will build the fixtures with IL2CPP and output universal APKs containing ARMv7, ARM64 and x86 slices.

This is important as this is how most users will build their projects too

## Changes

- Changed Scripting backend from Mono to IL2CPP
- Enabled ARM64 Architecture
- Updated project settings to match unity 2018 standards as we have dropped support for 2017 

## Testing

I manually tested this by changing the settings in the fixture and then building the fixture with the CI build script and inspecting the output. Then i ran a few android tests locally.

### DIFF
--- a/features/android/android_csharp_errors.feature
+++ b/features/android/android_csharp_errors.feature
@@ -16,10 +16,8 @@ Feature: Android smoke tests for C# errors
         And the event "severityReason.type" equals "handledException"
 
         # Stacktrace validation
-        And the error payload field "events.0.exceptions.0.stacktrace" is a non-empty array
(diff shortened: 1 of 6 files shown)


twometresteve:
Raw project settings don't mean too much to me and more detail in the description would be useful.   What can we do after this change that we couldn't do beforehand?  What testing has been performed to show the change is correct?  That doesn't necessarily have to be automated tests, but having an understanding of what the problem was and that this solves that problem would be good.

twometresteve (approved):

PR #500, opened 2022-01-21
Bump bugsnag-android to v5.19.1

lemnik:
## Goal
Bumps bugsnag-android dependency to v5.19.1.

Note: that as bugsnag-android now compiles with Kotlin 1.4, it might end up being necessary to update the Kotlin stdlib jar


rich-bugsnag (approved):
thanks!

PR #503, opened 2022-02-04
Skip Android e2e tests on Unity 2021 pending further investigation

twometresteve:
## Goal

Skip Android e2e tests on Unity 2021 pending further investigation.


rich-bugsnag (approved):
thanks

PR #505, opened 2022-02-07
fix(featureFlags): replace variant if flag exists

rich-bugsnag:
## Goal

My original implementation did not replace the variant of an existing feature.

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -585,18 +585,25 @@ namespace BugsnagUnity
 
         public void AddFeatureFlag(string name, string variant = null)
         {
-            _featureFlags.Add(new FeatureFlag(name, variant));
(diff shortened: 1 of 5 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -585,18 +585,25 @@ public void SetUser(string id, string email, string name)
...
         public void AddFeatureFlags(FeatureFlag[] featureFlags)
         {
-            _featureFlags.AddRange(featureFlags);
             foreach (var flag in featureFlags)
             {
-                NativeClient.AddFeatureFlag(flag.Name, flag.Variant);
+                AddFeatureFlag(flag.Name, flag.Variant);
             }
-        }
+        }
```
nickdowell, line 606: 🤨

nickdowell (approved):
Looks good apart from some minor whitespace and formatting issues

PR #509, opened 2022-02-16
task(persistence): implement persisting sessions when delivery fails

rich-bugsnag:
## Goal

implement persisting sessions when delivery fails

## Design

The flow is as follows.
1: When a session is recorded it gets added to a list of pending session payloads in memory
2: If the session is sent successfully then the pending session is removed
3: If the sessio fails to send it is written to disk, then on next startup or when the app comes back into focus we try and send it again

## Changeset

- Added FileManager class
- Added config value MaxPersistedSessions and pass the value through to Android and Cocoa
- Display MaxPersistedSessions in the bugsnag config unity window
- Try sending cached payloads on launch and when app regains focus

## Testing

Added CI tests for Windows and WebGL


nickdowell (commented):
Was persisting sessions prior to sending a deliberate design decision?

We not too long ago changed the Cocoa notifier to only persist sessions if sending failed, in order to reduce disk activity...

nickdowell (approved):

PR #511, opened 2022-02-24
Tms/attach requests

twometresteve:
## Goal

Upload requests received in all e2e test scenarios as build artifacts, pass or fail.

## Design

Maze Runner generates these files as standard, separated into passed and failed folders.  It's just a matter of selecting which file to upload.

## Changeset

Global find/replace of the existing file pattern, which uploaded files for failed scenarios only.  Also moved a couple of steps that had been missed over to our new build servers.

## Testing

Covered by a full CI run - see the artifacts tab for each Maze Runner step.


rich-bugsnag (approved):
Love it, thanks!

PR #512, opened 2022-02-25
task(persistence): persist events

rich-bugsnag:
## Goal

Persist csharp events if they fail to send

## Design

The feature works as follows:

1: When an event is ready to send, it does through any on send error callbacks and unless discarded by a callback gets packaged into a Report

2: That report gets added to a list of pending reports

3: If the report sends correctly then it is removed from the list of pending reports

4: If the report fails with a 408, 429 or 500+ error code then we wait and try again, if it fails with another 400 code then we discard it.

5: If the report fails with another error code then it gets serialised and written to disk

6: On the next app launch or regain of focus, we try to send any cached reports

7: Each cached report gets deserialised back into a Report object and the event within gets run through any registered OnSendError callbacks

8: If not discarded by a callback, the report trys to send again, if it sends, the cached report is removed, if it fails the cached report is updated

## Changeset

- Added methods to FileManager.cs to handle event reports
- Added the ability to deserialise the following classes from jason to allow them to be accesed in callbacks: App. AppWithState, Device, DeviceWithState, Breadcrumb, Error, Event, FeatureFlag, Report, StacktraceLine
- Added config option MaxPersistedEvents

## Testing

Added full CI tests

### DIFF
--- a/features/desktop/persistance.feature
+++ b/features/desktop/persistance.feature
@@ -61,7 +61,7 @@ Feature: Unity Persistance
         And the event "breadcrumbs.0.name" equals "Persist Message"
 
         #metadata
-        And the event "metaData.Persist.Persist" equals "Persist"
+        And the event "metaData.Persist Section.Persist Key" equals "Persist Value"
         And I wait for 5 seconds
 
 
--- a/features/desktop/persistance.feature
+++ b/features/desktop/persistance.feature
@@ -1,0 +1,66 @@
+Feature: Unity Persistance
+
+    @skip_webgl
+    Scenario: Receive a persisted session mac and windows
+        When I run the game in the "PersistSession" state
+        And I wait for 6 seconds
+        And I run the game in the "PersistSessionReport" state
+        And I wait to receive 2 sessions
+        Then the session is valid for the session reporting API version "1.0" for the "Unity Bugsnag Notifier" notifier
+        And the session payload field "app.releaseStage" equals the platform-dependent string:
+            | macos | Second Session |
+            | windows | First Session |
+        And I discard the oldest session
+        And the session payload field "app.releaseStage" equals the platform-dependent string:
+            | macos | First Session |
+            | windows | Second Session |
+
+    @webgl_only
+    Scenario: Receive a persisted session webgl
+        When I run the game in the "PersistSession" state
+        And I wait for 10 seconds
+        And I run the game in the "PersistSessionReport" state
+        And I wait to receive 2 sessions
+        Then the session is valid for the session reporting API version "1.0" for the "Unity Bugsnag Notifier" notifier
+        And the session payload field "app.releaseStage" equals "First Session"
+        And I discard the oldest session
+        And the session payload field "app.releaseStage" equals "Second Session"
+        And I wait for 5 seconds
+
+    Scenario: Receive a persisted event
+        When I run the game in the "PersistEvent" state
+        And I wait for 10 seconds
+        And I run the game in the "PersistEventReport" state
+        And I wait to receive 2 errors
+        And the event "context" equals "First Error"
+        And the exception "message" equals "First Event"
+        And I discard the oldest error
+        And the event "context" equals "Second Error"
+        And the exception "message" equals "Second Event"
+
+    Scenario: Receive a persisted event with on send callback
+        When I run the game in the "PersistEvent" state
+        And I wait for 10 seconds
+        And I run the game in the "PersistEventReportCallback" state
+        And I wait to receive 2 errors
+        And the event "context" equals "First Error"
+        And the exception "message" equals "First Event"
+
+        #device metadata
+        And the event "device.id" equals "Persist Id"
+
+        #app metadata
+        And the event "app.binaryArch" equals "Persist BinaryArch"
+ 
+        #exception data
+        And the event "exceptions.0.errorClass" equals "Persist ErrorClass"
+        And the event "exceptions.0.stacktrace.0.method" equals "Persist Method"
+
+        #breadcrumbs
+        And the event "breadcrumbs.0.name" equals "Persist Message"
+
+        #metadata
+        And the event "metaData.Persist.Persist" equals "Persist"
+
+
+        
--- a/features/desktop/persistance.feature
+++ b/features/desktop/persistance.feature
@@ -37,6 +37,7 @@ Feature: Unity Persistance
         And I discard the oldest error
         And the event "context" equals "Second Error"
         And the exception "message" equals "Second Event"
+        And I wait for 5 seconds
 
     Scenario: Receive a persisted event with on send callback
         When I run the game in the "PersistEvent" state
@@ -61,6 +62,7 @@ Feature: Unity Persistance
 
         #metadata
         And the event "metaData.Persist.Persist" equals "Persist"
+        And I wait for 5 seconds
 
 
         
--- a/features/fixtures/maze_runner/Assets/Scripts/Main.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Main.cs
@@ -159,7 +159,7 @@ public class Main : MonoBehaviour
         {
             case "PersistEvent":
                 config.AutoDetectErrors = true;
-                config.Endpoints = new EndpointConfiguration("https://notify.bugsdnag.com", "https://notify.bugsdnag.com");
+                config.Endpoints = new EndpointConfiguration("https://notify.def-not-bugsnag.com", "https://notify.def-not-bugsnag.com");
                 config.Context = "First Error";
                 break;
             case "PersistEventReport":
--- a/features/fixtures/maze_runner/Assets/Scripts/Main.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Main.cs
@@ -185,7 +185,7 @@ public class Main : MonoBehaviour
                         crumb.Message = "Persist Message";
                     }
 
-                    @event.AddMetadata("Persist", new Dictionary<string, object> { { "Persist", "Persist" } });
+                    @event.AddMetadata("Persist Section", new Dictionary<string, object> { { "Persist Key", "Persist Value" } });
 
                     return true;
                 });
--- a/features/fixtures/maze_runner/Assets/Scripts/Main.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Main.cs
@@ -157,6 +157,39 @@ public class Main : MonoBehaviour
     {
         switch (scenario)
         {
+            case "PersistEvent":
+                config.AutoDetectErrors = true;
+                config.Endpoints = new EndpointConfiguration("https://notify.bugsdnag.com", "https://notify.bugsdnag.com");
+                config.Context = "First Error";
+                break;
+            case "PersistEventReport":
+                config.AutoDetectErrors = true;
+                config.Context = "Second Error";
+               
+                break;
+            case "PersistEventReportCallback":
+                config.AutoDetectErrors = true;
+                config.Context = "Second Error";
+                config.AddOnSendError((@event) => {
+
+                    @event.App.BinaryArch = "Persist BinaryArch";
+
+                    @event.Device.Id = "Persist Id";
+
+                    @event.Errors[0].ErrorClass = "Persist ErrorClass";
+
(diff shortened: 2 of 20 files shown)


src/BugsnagUnity/Payload/SessionReport.cs
```diff
@@ -44,9 +44,22 @@ public SessionReport(Configuration configuration, Dictionary<string, object> cac
...
-            this.AddToPayload("sessions", new [] { cachedReport.Get("session") });
+            this.AddToPayload("sessions", cachedReport.Get("sessions"));
             Id = cachedReport["payloadId"].ToString();
         }
 
+        public Dictionary<string, object> GetSerialisableSessionReport()
+        {
+            var serialisableSessionReport = new Dictionary<string, object>
+                {
+                    { "payloadId", Id },
```
kattrali, line 55: Session has `payloadId` but event has `id` - is that intended?

features/desktop/persistance.feature
```diff
@@ -0,0 +1,68 @@
+Feature: Unity Persistance
```
gareththackeray, line 1: Could you change "persistance" to "persistence" everywhere?  Including filenames?

src/BugsnagUnity/FileManager.cs
```diff
@@ -106,30 +128,64 @@ internal static void CacheSession(string reportId)
...
+            using (var stream = new MemoryStream())
+            using (var reader = new StreamReader(stream))
+            using (var writer = new StreamWriter(stream, new UTF8Encoding(false)) { AutoFlush = false })
+            {
+                SimpleJson.SerializeObject(eventReport.GetSerialisableEventReport(), writer);
+                writer.Flush();
+                stream.Position = 0;
+                var jsonString = reader.ReadToEnd();
+                var path = EventsDirectory + "/" + eventReport.Id + ".event";
+                WriteToDisk(jsonString, path);
```
gareththackeray, line 147: Wouldn't it make more sense to stream straight to disk rather than streaming into a string and then writing to disk?

src/BugsnagUnity/FileManager.cs
```diff
@@ -106,30 +128,64 @@ internal static void CacheSession(string reportId)
...
+        private static void RemoveOldestFile(string[] filePaths)
         {
-            var oldestSession = string.Empty;
+            var oldestFilePath = string.Empty;
             var oldestMillis = default(double);
-            foreach (var cachedSessionPath in _cachedSessions)
+            foreach (var filePath in filePaths)
             {
-                var milliesSinceCreated = (DateTime.UtcNow - File.GetCreationTimeUtc(cachedSessionPath)).TotalMilliseconds;
+                var milliesSinceCreated = (DateTime.UtcNow - File.GetCreationTimeUtc(filePath)).TotalMilliseconds;
```
gareththackeray, line 178:
```suggestion
                var millisSinceCreated = (DateTime.UtcNow - File.GetCreationTimeUtc(filePath)).TotalMilliseconds;
```

src/BugsnagUnity/FileManager.cs
```diff
@@ -106,30 +128,64 @@ internal static void CacheSession(string reportId)
...
                 filesCount = _cachedSessions.Length;
             }
         }
 
-        private static void RemoveOldestSession()
+        private static void RemoveOldestFile(string[] filePaths)
         {
-            var oldestSession = string.Empty;
+            var oldestFilePath = string.Empty;
             var oldestMillis = default(double);
```
gareththackeray, line 175: Minor and unrelated to this change but if you want 0 I think it's better to say 0!

src/BugsnagUnity/FileManager.cs
gareththackeray, line 178: Again unrelated to this change but if we're looking for the oldest creation time why don't we just compare the datetime rather than doing a calculation on it first?

E.g. `if (File.GetCreationTimeUtc(filePath)) < oldestFileDateTime)` (and initialise `oldestFileDateTime` to `DateTime.UtcNow` (or `DateTime.Max`).

Alternatively we could use `OrderBy()` and then remove the first / last.  If you ordered by at the time you populated the field you wouldn't need to keep searching for the oldest one as you'd know.

src/BugsnagUnity/FileManager.cs
```diff
@@ -106,30 +128,64 @@ internal static void CacheSession(string reportId)
...
             while(filesCount > _configuration.MaxPersistedSessions)
             {
-                RemoveOldestSession();
+                RemoveOldestFile(_cachedSessions);
                 filesCount = _cachedSessions.Length;
             }
         }
 
-        private static void RemoveOldestSession()
+        private static void RemoveOldestFile(string[] filePaths)
```
gareththackeray, line 172: It doesn't look like we actually remove the entry from `_cachedSessions` or `_cachedEvents` after deletion.  Is this important?  Is it possible that this could be called twice without them being repopulated?

src/BugsnagUnity/FileManager.cs
```diff
@@ -141,10 +197,28 @@ internal static void PayloadSendSuccess(IPayload payload)
                     RemovedCachedSession(payload.Id);
                     break;
                 case PayloadType.Event:
+                    RemovedCachedEvent(payload.Id);
+                    RemovePendingEvent(payload.Id);
                     break;
             }
         }
 
+        internal static void RemovedCachedEvent(string id)
```
gareththackeray, line 206: Should this be `Removed` or just `Remove`?    Same for the session equivalent.

src/BugsnagUnity/IDelivery.cs
```diff
@@ -116,25 +116,41 @@ IEnumerator PushToServer(IPayload payload, byte[] body)
...
+        }
+
+        private bool IsRetryableNow(long code)
+        {
+            return code >= 500 || code == 408 || code == 429;
+        }
+
+        private bool ShouldCache(long code)
+        {
+            return code < 400;
```
gareththackeray, line 151: When would this ever happen?  Should this be `< 500`?

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -157,6 +157,39 @@ void PrepareConfigForScenario(Configuration config, string scenario)
...
+                    @event.Errors[0].ErrorClass = "Persist ErrorClass";
+
+                    @event.Errors[0].Stacktrace[0].Method = "Persist Method";
+
+                    foreach (var crumb in @event.Breadcrumbs)
+                    {
+                        crumb.Message = "Persist Message";
+                    }
+
+                    @event.AddMetadata("Persist", new Dictionary<string, object> { { "Persist", "Persist" } });
```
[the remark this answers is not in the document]
rich-bugsnag: Done 👍

features/desktop/persistance.feature
rich-bugsnag: done 👍

src/BugsnagUnity/Payload/SessionReport.cs
rich-bugsnag: changed to id 👍

src/BugsnagUnity/FileManager.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/FileManager.cs
rich-bugsnag: done 👍

src/BugsnagUnity/IDelivery.cs
rich-bugsnag: It makes sense in the context of the code handling the request. This is the last check after we have alread processed 200s 400s and 500s. So basically, if after all that, there is any error that's not a 400, like a 0 for example, then we should cache and try on next launch because it's most likely some kind of network issue.

src/BugsnagUnity/FileManager.cs
rich-bugsnag: They are repopulated whenever accessed: 
`        private static string[] _cachedSessions => Directory.GetFiles(SessionsDirectory, "*.session");
`

gareththackeray: oh I got confused by the `_` prefix.  That's normally used for fields rather than properties.

gareththackeray: PascalCase would be usual for a property.

src/BugsnagUnity/IDelivery.cs
gareththackeray: We haven't processed "400s" though we've specifically processed 408 and 429.

gareththackeray: Ah ok.  There's not really any such thing as ResponseCode 0; I guess it's just a convention.  There's a `response` property on the `UnityWebRequest` I wonder if we could use that instead, just to be clear.

Even if we don't do that, I'd suggest we specifically check for `0` here as it'd be much clearer what the intent is.

src/BugsnagUnity/FileManager.cs
```diff
@@ -141,10 +197,28 @@ internal static void PayloadSendSuccess(IPayload payload)
...
+                }
+            }
+        }
+
+        private static void RemovePendingEvent(string id)
+        {
+            _pendingEvents.RemoveAll(item => item.Id == id);
+        }
+
         internal static void RemovedCachedSession(string id)
```
gareththackeray, line 222: Should be `Remove` right?

src/BugsnagUnity/FileManager.cs
rich-bugsnag: Done 👍

src/BugsnagUnity/FileManager.cs
rich-bugsnag: Rewrote the method with OrderBy 👍

src/BugsnagUnity/FileManager.cs
```diff
@@ -164,27 +147,17 @@ private static void CheckForMaxCachedSessions()
...
-                var milliesSinceCreated = (DateTime.UtcNow - File.GetCreationTimeUtc(filePath)).TotalMilliseconds;
-                if (milliesSinceCreated > oldestMillis)
-                {
-                    oldestMillis = milliesSinceCreated;
-                    oldestFilePath = filePath;
-                }
-            }
-            if (!string.IsNullOrEmpty(oldestFilePath))
+            var ordered = filePaths.OrderBy(file => File.GetCreationTimeUtc(file)).ToArray();
+            for (int i = 0; i < numToRemove; i++)
```
gareththackeray, line 158: You could also have done:

```
foreach(var file in ordered.Take(numToRemove))
{
  File.Delete(file);
}
```

src/BugsnagUnity/FileManager.cs
```diff
@@ -284,17 +257,21 @@ internal static List<IPayload> GetCachedPayloads()
...
-            File.WriteAllText(path, json);
+            using var fileStream = File.Create(path);
+            using var writer = new StreamWriter(fileStream, new UTF8Encoding(false)) { AutoFlush = false };
+            SimpleJson.SerializeObject(data, writer);
         }
 
-        private static void CheckForDirectoryCreation()
+
+
+private static void CheckForDirectoryCreation()
```
gareththackeray, line 274: indendentation gone weird

src/BugsnagUnity/FileManager.cs
```diff
@@ -11,21 +12,30 @@ public class FileManager
...
         }
 
+        private static string EventsDirectory
+        {
+            get { return CacheDirectory + "/Events"; }
+        }
+
         private static string[] _cachedSessions => Directory.GetFiles(SessionsDirectory, "*.session");
 
+        private static string[] _cachedEvents => Directory.GetFiles(EventsDirectory, "*.event");
```
gareththackeray, line 36: C# norms + consistency with other properties in this file.

```suggestion
        private static string[] CachedSessions => Directory.GetFiles(SessionsDirectory, "*.session");

        private static string[] CachedEvents => Directory.GetFiles(EventsDirectory, "*.event");
```

src/BugsnagUnity/FileManager.cs
rich-bugsnag: done 👍

kattrali (approved):

PR #513, opened 2022-02-28
Exit script if any command fails

twometresteve:
## Goal

Cleans up some scripts by simply using the `-e` option to exit a script if any individual command fails.

## Testing

Covered by CI.


rich-bugsnag (approved):
Thanks 👍

PR #516, opened 2022-03-02
fix(CI): clear cache after persistence tests

rich-bugsnag:
## Goal

clear cache after persistence tests to stop artifacts causing false results

## Testing

Full CI Run


twometresteve (approved):
I think there's a slight more elegant and generic way of doing this in the Cucumber, but if it fixes the current flake then I'm on board and can make a slight adjustment myself at a later stage.

PR #518, opened 2022-03-10
task(persistence): bring persistence up to spec

rich-bugsnag:
## Goal

- Remove legacy delivery queue
- Modifications to the event in OnSend should not be persisted
- Process the cached events in the background, one by one

## Changeset

- Removed the IDelivery interface and it's inclusion and initialisation in the NativeInterface class as it was unnecessary. 
- Replaced it with a simpler Delivery class that is initialised in the Client.
- Move the processing of OnSendError callbacks to the Delivery class so that they run after the event payload has been stored for potential caching.
- Removed the delayed retry of payload sending as this is now handled by caching and retrying later.
- Simplified the Filemanager class to allow for generic serialisation of pending requests before they go through callbacks.
- Process the cached events in the background, one by one using a coroutine

## Testing

Changes covered by existing tests.

### DIFF
--- a/features/desktop/persistence.feature
+++ b/features/desktop/persistence.feature
@@ -88,4 +88,20 @@ Feature: Unity Persistence
         And the error payload field "events.0.device.id" equals the stored value "device_id"
         And the error payload field "events.0.user.id" equals the stored value "device_id"
 
+    Scenario: Max Persisted Events
+        When I run the game in the "ClearBugsnagCache" state
+        And I wait for 5 seconds
+        And I run the game in the "MaxPersistEvents" state
+        And I wait for 5 seconds
+        And I run the game in the "(noop)" state
+        And I wait to receive 4 errors
+        And the exception "message" equals "Event 1"
+        And I discard the oldest error
+        And the exception "message" equals "Event 2"
+        And I discard the oldest error
+        And the exception "message" equals "Event 3"
+        And I discard the oldest error
+        And the exception "message" equals "Event 4"
+
+
         
--- a/features/desktop/persistence.feature
+++ b/features/desktop/persistence.feature
@@ -38,11 +38,11 @@ Feature: Unity Persistence
         And I wait for 5 seconds
         And I run the game in the "PersistEventReport" state
         And I wait to receive 2 errors
-        And the event "context" equals "First Error"
-        And the exception "message" equals "First Event"
-        And I discard the oldest error
         And the event "context" equals "Second Error"
         And the exception "message" equals "Second Event"
+        And I discard the oldest error
+        And the event "context" equals "First Error"
+        And the exception "message" equals "First Event"
 
 
     Scenario: Receive a persisted event with on send callback
@@ -52,6 +52,7 @@ Feature: Unity Persistence
         And I wait for 5 seconds
         And I run the game in the "PersistEventReportCallback" state
         And I wait to receive 2 errors
+        And I discard the oldest error
         And the event "context" equals "First Error"
         And the exception "message" equals "First Event"
 
--- a/src/BugsnagUnity/Delivery.cs
+++ b/src/BugsnagUnity/Delivery.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Text;
@@ -15,6 +16,11 @@ namespace BugsnagUnity
         private GameObject _dispatcherObject;
         private Configuration _configuration;
         private object _callbackLock { get; } = new object();
+
+        private static List<string> _finishedCacheDeliveries = new List<string>();
+
+        private bool _cacheDeliveryInProcess;
+        private WaitForSeconds _deliverCacheWaitTime = new WaitForSeconds(0.1f);
 
         internal Delivery(Configuration configuration)
         {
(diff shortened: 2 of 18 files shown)


src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        private IEnumerator DeliverCachedPayloads(string[] payloadPaths)
+        {
+            foreach (var cachedPayloadPath in payloadPaths)
+            {
+                var payload = FileManager.GetPayloadFromCachePath(cachedPayloadPath);
+                if (payload != null)
+                {
+                    Send(payload);
+                    while (!CachedPayloadProcessed(payload.Id))
```
tomlongridge, line 162: Can you use [WaitUntil](https://docs.unity3d.com/ScriptReference/WaitUntil.html) checking the processed list for this rather than our own polling?

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,172 @@
...
+            if (_cacheDeliveryInProcess)
+            {
+                return;
+            }
+            _cacheDeliveryInProcess = true;
+            try
+            {
+                _finishedCacheDeliveries.Clear();
+                var payloads = FileManager.GetCachedPayloadPaths();
+                MainThreadDispatchBehaviour.Instance().Enqueue(DeliverCachedPayloads(payloads));
```
tomlongridge, line 134: The other use of `MainThreadDispatchBehaviour` calls `CreateDispatchBehaviour` - do we need the equivalent here? Do these two separate calls to the dispatcher play nicely together?

src/BugsnagUnity/FileManager.cs
```diff
@@ -246,74 +234,40 @@ internal static void RemoveCachedSession(string id)
...
+                if (path.Contains("session"))
                 {
-                    RemoveCachedEvent(eventReportFromCachedPayload.Id);
-                    RemovePendingEvent(eventReportFromCachedPayload.Id);
+                    var json = File.ReadAllText(path);
+                    var sessionReportFromCachedPayload = new SessionReport(_configuration, ((JsonObject)SimpleJson.DeserializeObject(json)).GetDictionary());
+                    return sessionReportFromCachedPayload;
                 }
-                else
+                if (path.Contains("event"))
```
tomlongridge, line 255:
```suggestion
                else if (path.Contains("event"))
```

src/BugsnagUnity/FileManager.cs
```diff
@@ -246,74 +234,40 @@ internal static void RemoveCachedSession(string id)
...
-                            break;
-                        }
-                    }
-                    catch
-                    {
-                        // If the callback causes an exception, ignore it and execute the next one
-                    }
-                }
-                if (shouldDiscard)
+                if (path.Contains("session"))
```
tomlongridge, line 249: Suggest making a `const` out of ".session" and ".event" and testing against that rather than string values to make it clear what you're looking for. Presumably this could also be an `EndsWith`?

rich-bugsnag: done 👍

src/BugsnagUnity/FileManager.cs
rich-bugsnag: done 👍

src/BugsnagUnity/Delivery.cs
rich-bugsnag: Added the check for it's existence 👍

src/BugsnagUnity/Delivery.cs
rich-bugsnag: nice, done 👍

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,172 @@
...
+                    // not avaliable in unit tests
+                    MainThreadDispatchBehaviour.Instance().Enqueue(PushToServer(payload, body));
+                }
+                catch { }
+            }
+        }
+
+        public void Send(IPayload payload)
+        {
+            if (payload.PayloadType == PayloadType.Event)
```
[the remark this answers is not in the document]
rich-bugsnag: done 👍

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,172 @@
...
+                if (code == 200 || code == 202)
+                {
+                    // success!
+                    FileManager.PayloadSendSuccess(payload);
+                }
+                else if (code >= 500 || code == 408 || code == 429 || req.isNetworkError || code < 400)
+                {
+                    // sending failed, cache payload to disk
+                    FileManager.SendPayloadFailed(payload);
+                }
```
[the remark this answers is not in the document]
rich-bugsnag: Done 👍

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            if (_cacheDeliveryInProcess)
+            {
+                return;
+            }
+            _cacheDeliveryInProcess = true;
+            try
+            {
+                _finishedCacheDeliveries.Clear();
+                var payloads = FileManager.GetCachedPayloadPaths();
```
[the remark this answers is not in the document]
rich-bugsnag: done 👍

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        private static List<string> _finishedCacheDeliveries = new List<string>();
+
+        private bool _cacheDeliveryInProcess;
+
+
+        internal Delivery(Configuration configuration)
+        {
+            _configuration = configuration;
+            CreateDispatchBehaviour();
```
tomlongridge, line 28: If this is done in the constructor (like it is in master), I don't see a reason that this needs to be called just before use? This could just be inlined:

```suggestion
            _dispatcherObject = new GameObject("Bugsnag main thread dispatcher");
            _dispatcherObject.AddComponent<MainThreadDispatchBehaviour>();
```

src/BugsnagUnity/Delivery.cs
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            if (_dispatcherObject == null)
+            {
+                CreateDispatchBehaviour();
+            }
+            try
+            {
+                // not avaliable in unit tests
+                MainThreadDispatchBehaviour.Instance().Enqueue(PushToServer(payload, SerializePayload(payload)));
```
tomlongridge, line 68: Does this mean that the serialisation won't be done in the thread?

features/desktop/persistence.feature
```diff
@@ -88,4 +88,20 @@ Feature: Unity Persistence
...
+        And I wait for 5 seconds
+        And I run the game in the "(noop)" state
+        And I wait to receive 4 errors
+        And the exception "message" equals "Event 1"
+        And I discard the oldest error
+        And the exception "message" equals "Event 2"
+        And I discard the oldest error
+        And the exception "message" equals "Event 3"
+        And I discard the oldest error
+        And the exception "message" equals "Event 4"
```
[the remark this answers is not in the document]
tomlongridge: Discussing with @rich-bugsnag and we think order here is important after all, so happy to leave it as it was. (Apologies for getting you to dig that out @Cawllec!)

tomlongridge (approved):

PR #522, opened 2022-03-16
fix(CI): give requests more time to fix ordering issues

rich-bugsnag:
## Goal

The test features/desktop/persistence.feature:91 was flaky because it depends on requests completing in order, and some requests take longer than others, even when there is the same response.

## Changeset

- Added a 2 second wait between each request to allow the request to complete in the correct order.
- Added a 5 second wait before checking for the 4 expected requests to make sure that only 4 come through.


tomlongridge (approved):
It's not elegant, but I follow your logic!

I guess we could shave 4 seconds off by only notifying 3 times? 1st one is dropped; 2nd and 3rd and are sent; the 4th and 5th don't really add much to the test.

PR #527, opened 2022-03-18
Add automatic retry to all pipeline steps

joshedney:
## Goal

Add automatic retry to the build steps in the pipeline to automatically retry potential flakey build steps.

## Changeset

add the following code snippet to each build step within the pipeline file:

```yaml
retry:
  automatic:
    - exit_status: "*"
      limit: 1
```
This will retry a step if it has an exit status of anything other than `0`

## Testing

Full CI has been run.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -85,10 +85,6 @@ steps:
           - Mazerunner.log
     commands:
       - scripts/ci-run-macos-tests.sh
-    retry:
-      automatic:
-        - exit_status: "*"
-          limit: 1
 
   - label: Run MacOS e2e tests for Unity 2019
     timeout_in_minutes: 30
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -272,6 +284,10 @@ steps:
     concurrency: 9
     concurrency_group: browserstack-app
     concurrency_method: eager
+    retry:
+      automatic:
+        - exit_status: "*"
+          limit: 1
```
twometresteve, line 290: I wouldn't let test steps retry automatically, as it could mask flakes that need investigations.  Retries for things within the step that we're happy to be retried (e.g. app uploads) is the safest approach.

twometresteve: ...retrying build steps seems reasonable, as we can't control any flakiness in the Unity build system.

joshedney: That makes sense

twometresteve (approved):

PR #528, opened 2022-03-21
task(edm): add optional edm support

rich-bugsnag:
## Goal

Add a menu item that toggles support for EDM4U android builds.

## Design

When the user clicks the “Enable EDM Support“ menu item 2 things will happen:

1: Our editor code will create a file with the above xml named BugsnagAndroidDependencies.xml in Bugsnag/Editor.

This will allow EDM to recognise it as an EDM manifest and import the resolve the correct libs.

2: Our editor code will change the import settings of our bundled Kotlin libs, telling the Unity editor to ignore them. 

3: A log message will notify the user if the process was a success or if there were any issues.

Disabling EDM Support
When the user clicks the “Disabling EDM Support“ menu item 2 things will happen:

1: Our editor code will delete the xml dependency manifest (if it exists);

2: Our editor code will change the import settings of our bundled Kotlin libs, telling the unity editor to include them. 

3: A log message will notify the user if the process was a success or if there were any issues.

## Changeset

- Moved the kotlin specific jars to a dedicated kotlin folder for easier manipulation.
- Added Tickbox menu item to enable/disable EDM support and log the result, see BugsnagEditor.cs for details
- Added CI testing fixture and build scripts for future CI testing

## Testing

Will be manually tested for now with planned CI support coming soon.


src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -6,12 +6,21 @@
...
 using UnityEditor.Callbacks;
 using System.Linq;
+using System;
 
 namespace BugsnagUnity.Editor
 {
     public class BugsnagEditor : EditorWindow
     {
 
+        private const string ANDROID_DEPS_XML = "<dependencies><androidPackages><repositories><repository>https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib</repository></repositories><androidPackage spec=\"org.jetbrains.kotlin:kotlin-stdlib:1.5.0\"></androidPackage></androidPackages></dependencies>";
```
tomlongridge, line 16: I believe Android is on v1.4.32 of `kotlin-stdlib` - see [Versions.kt](https://github.com/bugsnag/bugsnag-android/blob/next/buildSrc/src/main/kotlin/com/bugsnag/android/Versions.kt#L14) and the Rakefile for Unity uses v1.3.72.

Are we deliberately targeting a newer version?

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -6,12 +6,21 @@
...
+
+        private const string EDM_MENU_ITEM = "Window/Bugsnag/Enable EDM4U Support";
+
+        private static string EDMDepsFilePath = "/Bugsnag/Editor/BugsnagAndroidDependencies.xml";
+
+        private static string KotlinLibsDirPath = "/Bugsnag/Plugins/Android/Kotlin";
```
tomlongridge, line 22: These are always used with `Application.dataPath` - suggest prepending it here.

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -224,5 +253,125 @@ public static void OnPostProcessBuild(BuildTarget target, string path)
...
+
+                File.WriteAllText(path, ANDROID_DEPS_XML);
+
+                foreach (var lib in GetKotlinLibs())
+                {
+                    lib.SetCompatibleWithPlatform(BuildTarget.Android, false);
+                    lib.SaveAndReimport();
+                }
+
+                AssetDatabase.Refresh();
```
tomlongridge, line 271: Suggest abstracting this to a `UpdateLibraryImportSettings(bool)` private method with it's own try-catch (if required) that can be shared with `DisableEDM` and make this function do fewer things.

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -224,5 +253,125 @@ public static void OnPostProcessBuild(BuildTarget target, string path)
...
+        }
+
+        private static bool IsEDMEnabled()
+        {
+            var success = File.Exists(Application.dataPath + EDMDepsFilePath);
+            foreach (var lib in GetKotlinLibs())
+            {
+                if (lib.GetCompatibleWithPlatform(BuildTarget.Android))
+                {
+                    success = false;
```
tomlongridge, line 348: You could short-circuit here for a small gain.

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
tomlongridge, line 16: Ideally we would have some link to the kotlin lib added by the Rakefile so that we end up with the same version and don't risk there being differences depending on EDM support.

At minimum this could be a comment in both places saying to update the other, unless you can see a more sophisticated approach to define the version in one place.

features/fixtures/EDM_Fixture/Assets/ExternalDependencyManager/Editor/LICENSE
```diff
@@ -0,0 +1,245 @@
+Copyright (C) 2014 Google Inc.
```
tomlongridge, line 1: Is it feasible to curl this dependency when we run the test rather than check it in? This makes the version we're testing against a little more explicit - potentially always testing against the [latest](https://github.com/googlesamples/unity-jar-resolver/blob/master/external-dependency-manager-latest.unitypackage)?

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
rich-bugsnag: You can't use Application.dataPath in an initialiser. Unity throws an error if you do.

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
rich-bugsnag: Done 👍

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
rich-bugsnag: Done 👍

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
rich-bugsnag: @lemnik got any insight on this? Should i be targeting 1.4.32?

features/fixtures/EDM_Fixture/Assets/ExternalDependencyManager/Editor/LICENSE
rich-bugsnag: yes i think that's a good idea, i'll make a note in the CI task for when we get it all setup.

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
lemnik: Yup. I think 1.4.32 is a better option.

rich-bugsnag: I updated the versions everywhere to use 1.4.32, and added comments regarding keeping them up to date

rich-bugsnag: Done

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -224,5 +253,135 @@ public static void OnPostProcessBuild(BuildTarget target, string path)
...
         }
 #endif
+
+        public static void EnableEDM()
+        {
+            try
+            {
+                EditDepsFile(true);
+                UpdateKotlinLibraryImportSettings(false);
+                AssetDatabase.Refresh();
```
tomlongridge, line 263: nit: I would move this Refresh call to `UpdateKotlinLibraryImportSettings` as presumably it is needed due to the library file change, which the function is responsible for.

tomlongridge (approved):
LGTM - one small suggestion

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
rich-bugsnag: it's also needed for the deps file creation

PR #532, opened 2022-03-23
task(handler): implement new exception handler

rich-bugsnag:
## Goal

Currently, our SDK listens for errors reported by the Unity debugger. 

This means we only get the following information:

string condition, string stackTrace, LogType logType

In Unity 2019 it is possible to insert our own logger before the Unity one, and therefore get access to the original exception object.

This would mean we can report more accurate error information.

For example, with our current solution we cannot correctly report inner exceptions, with this change we can.

## Design

On startup we detect the unity version. If it is 2019+ then we insert a handler that can access the original Exception object. 
We still listen to the Unity logger for unity Debug Logs (log, warning, error).

If we detect Unity 2018 then we don't attach the new handler and listen to the unity logger as normal.

## Changeset

- Added the private class `Client.BugsnagLogHandler`. This is the new handler. It checks, against rate limiters and config options if an exception should be sent and then sends or ignores it.
- Added unity version check to Client to decide which handler to use.
- Added util methods in `Error.cs` and `StackTraceLine.cs` to help format android java exceptions sent through the new handler.
- Added new constructor in `UnityLogMessage.cs` help compare and rate limit exceptions sent through the new handler using the existing logic.

## Testing

- Added new test for InnerExceptions that runs on unity 2019+
- Changed existing tests to look for both the old style of event data and the new.

## Review Notes

The main thing to look out for here is that apart from event data, there should be no change in notifier behaviour, all config options should be respected as before and user experience should be the same. We just get slightly different data in some cases, and more data in others.


src/BugsnagUnity/Payload/Error.cs
```diff
@@ -137,19 +137,44 @@ internal static Error FromSystemException(System.Exception exception, System.Dia
...
+                }
+                else
+                {
+                    // There was no Java exception description, so the Java class is the only content in
+                    // the message.
+                    errorClass = match.Groups["errorClass"].Value;
+                    message = "";
+                }
+                var lines = new StackTrace(exception.StackTrace, StackTraceFormat.AndroidJava).ToArray();
+                return new Error(errorClass, message, lines, HandledState.ForUnhandledException(), true);
```
kattrali, line 162: perhaps this logic could be contained in something to process class and message, rather than repeating the content of `FromUnityLogMessage()`? for example, both could have something like `(string, string) FormatClassAndMessage(className, rawMessage)`

src/BugsnagUnity/Payload/StackTraceLine.cs
```diff
@@ -45,6 +45,10 @@ internal StackTrace(string stackTrace, StackTraceFormat format)
                   : StackTraceLine.FromLogMessage(lines[i]);
                 if (frame != null)
                 {
+                    if (frame.Method.StartsWith("at "))
+                    {
+                        frame.Method = frame.Method.TrimStart('a', 't', ' ');
+                    }
```
kattrali, line 51: I'm not really sure what case this is handling, maybe one for the [unit tests](https://github.com/bugsnag/bugsnag-unity/blob/next/tests/BugsnagUnity.Tests/StackFrameParsingTests.cs)?

src/BugsnagUnity/Payload/Error.cs
rich-bugsnag: Done 👍  Didint want to adda package dependancy to the sdk so i didnt use a string Tuple, used an array instead

src/BugsnagUnity/Payload/StackTraceLine.cs
rich-bugsnag: Android stacktraces coming through the new handler had "at " at the beginning of each frame method. It's coverd in CI tests. Is this ok?

src/BugsnagUnity/Payload/Error.cs
```diff
@@ -137,19 +137,44 @@ internal static Error FromSystemException(System.Exception exception, System.Dia
         {
             var errorClass = exception.GetType().Name;
             var stackFrames = new System.Diagnostics.StackTrace(exception, true).GetFrames();
+           
 
-            StackTraceLine[] lines = null;
-
-            if (stackFrames != null && stackFrames.Length > 0)
+            if (errorClass == ANDROID_JAVA_EXCEPTION_CLASS)
```
nickdowell, line 142: The addition of Android-specific handling is a bit unexpected and wasn't called out in the PR description.

Is it worth a few comments to explain why this is required?

kattrali: There's a (brief) explanation around [line 193](https://github.com/bugsnag/bugsnag-unity/pull/532/files#diff-18e358b00cc5c047e63661436c8c943364cff776be9ea7f016248fe594038d81L193) - this is the duplicate bit that's getting consolidated

src/BugsnagUnity/Payload/Error.cs
rich-bugsnag: Pushed

src/BugsnagUnity/Payload/Error.cs
rich-bugsnag: I added another comment explaining the JVM exception in the main thread situation 👍

src/BugsnagUnity/Payload/Error.cs
```diff
@@ -224,6 +243,23 @@ public static Error FromUnityLogMessage(UnityLogMessage logMessage, System.Diagn
             }
         }
 
+        public static bool ShouldSend(System.Exception exception)
+        {
+            if (exception.StackTrace == null)
```
kattrali, line 248: Is it possible that an exception has an empty stack trace _and_ it has one of the excluded error classes? it seems possible that an exception matching `ANDROID_JAVA_EXCEPTION_CLASS` could accidentally be sent by this line.

rich-bugsnag: That condition was unnecessary, i have removed it 👍

kattrali (approved):

PR #535, opened 2022-03-31
Change Buildkite Queue Name

joshedney:
## Goal

Change the macos 12 arm queue name to one that only includes the unity ready machines.

## Changeset

Change main buildkite queue to `macos-12-arm-unity` 

## Testing

[Full CI]


twometresteve (approved):
This obviously works and I would merge it as it, but I would note that not all steps in the pipeline using MacOS 12 needed Unity installing on them.  The steps that just need a mac to run could be moved to the more general queue, potentially saving build time (but I would do that as a separate PR).

PR #538, opened 2022-04-02
Bump Unity 2020 and 2021 test versions

twometresteve:
## Goal

Bump Unity 2020 and 2021 test versions.

## Changeset

For test steps that only need a mac to run (and not Unity installed) I have adjusted the queue to be the more general one.

## Testing

Covered by a full CI run.


joshedney (approved):
LGTM!

PR #552, opened 2022-05-03
Exit build script immediately on error

twometresteve:
## Goal

Exit script immediately on error.

## Design

Without this the build script can silently fail, leading to confusing failures in later steps.  Failing the step also means that it can take advantage of retrying the build (which isn't allowed on test steps).

## Testing

Covered by CI.


joshedney (approved):
LGTM!

PR #558, opened 2022-05-20
Release v7.0.0

rich-bugsnag:
Release v7.0.0

New Exception Interceptor


UPGRADING.md
```diff
@@ -1,6 +1,18 @@
 Upgrading
 =========
 
+## 6.x to 7.x
+
+### The new Exception interceptor
+
+When building using Unity 2019+, the Bugsnag SDK now uses a new method to intercept uncaught C# exceptions. This allows us access to the original exception object, meaning more accurate exception data and full support for nested/inner exceptions.
```
gareththackeray, line 8: I think "inner exception" is the correct C# term
```suggestion
When building using Unity 2019+, the Bugsnag SDK now uses a new method to intercept uncaught C# exceptions. This allows us access to the original exception object, meaning more accurate exception data and full support for inner exceptions.
```

UPGRADING.md
```diff
@@ -1,6 +1,18 @@
...
+
+This means that uncaught C# exceptions previously reported will not group automatically in the Bugsnag dashboard with exceptions reported by the new notifier.
```
gareththackeray, line 10:
```suggestion
This means that uncaught C# exceptions reported by `bugsnag-unity` V7+ will not group automatically in the Bugsnag dashboard with the equivalent exception reported by older versions of the library. 
```

UPGRADING.md
```diff
@@ -1,6 +1,18 @@
...
 
+## 6.x to 7.x
+
+### The new Exception interceptor
+
+When building using Unity 2019+, the Bugsnag SDK now uses a new method to intercept uncaught C# exceptions. This allows us access to the original exception object, meaning more accurate exception data and full support for nested/inner exceptions.
+
+This means that uncaught C# exceptions previously reported will not group automatically in the Bugsnag dashboard with exceptions reported by the new notifier. 
+
+All errors will still be reported, but you may wish to pay extra attention to new exceptions that come through once upgrading and group them accordingly.
```
gareththackeray, line 12: > group them accordingly.

What do you mean by this?  I don't think there's anything that can be done?

UPGRADING.md
```diff
@@ -1,6 +1,18 @@
...
+
+Handled errors and native errors will remain grouped and are not effected by this upgrade.
```
gareththackeray, line 14:
```suggestion
Handled errors and native crashes will continue grouping in the same way and are not affected by this upgrade.
```

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
 # Changelog
 
+## 7.0.0 (2022-04-23)
+
+This version contains **breaking** changes, as bugsnag-unity has been updated to use a more advanced Exception interceptor and existing grouping between errors in the Bugsnag dashboard may be broken.
```
tomlongridge, line 5: "broken" suggests not working, whereas we mean it has changed. How about:

```suggestion
This version contains **breaking** changes in the grouping of errors on the Bugsnag dashboard due to the introduction of a new exception interceptor that directly reports the thrown C# exceptions, rather than obtaining this information from Unity's logging mechanism. 
```

UPGRADING.md
```diff
@@ -1,6 +1,18 @@
 Upgrading
 =========
 
+## 6.x to 7.x
+
+### The new Exception interceptor
```
tomlongridge, line 6: I know `Exception` is a class name, but it irks me when I read it capitalised - in fact, we probably don't even need the h3 as it's the only noteworthy change.

UPGRADING.md
tomlongridge, line 10: Should we add a bit more info here as it doesn't really say _why_ it won't group.

e.g.

> Due to differences in the way types were previously exposed in Unity's logging mechanism and the type information available when the exception is intercepted, errors will be reported subtly differently once your app is upgraded to this version.

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+
+Please see the [upgrade guide](./UPGRADING.md) for details of all the changes and instructions on how to upgrade.
+
+In addition to the changes mentioned in the upgrade guide, the bundled Bugsnag Android Notifier has been updated. See below for details.
+
+* Update bugsnag-android to v5.22.3
```
tomlongridge, line 11: Maybe a good time to adopt a more brief syntax for this?

https://github.com/bugsnag/bugsnag-unreal/blob/next/CHANGELOG.md?plain=1#L13-L14

UPGRADING.md
gareththackeray: Updated my suggestion.

CHANGELOG.md
gareththackeray:
```suggestion
This version contains **breaking** changes: some errors generated by the new version will no longer group in the Bugnsag dashboard with equivalent errors generated by older versions. This is due to the introduction of a new exception interceptor that intercepts the C# exceptions objects, rather than parsing the information from Unity logs.
```

UPGRADING.md
rich-bugsnag: Yeah this is not worded correctly, i suppose i thought we should mention that you can group by context or by hash, but perhaps it's not necessary to say that

CHANGELOG.md
rich-bugsnag: Done 👍

UPGRADING.md
gareththackeray:
```suggestion
The impact of this will be that some errors will effectively be duplicated between older and newer versions of your app.
```

UPGRADING.md
```diff
@@ -1,6 +1,16 @@
...
 
+## 6.x to 7.x
+
+When building using Unity 2019+, the Bugsnag SDK now uses a new method to intercept uncaught C# exceptions. This allows us access to the original exception object, meaning more accurate exception data and full support for inner exceptions.
+
+This has slightly changed the format used for method names and signatures, which means that uncaught C# exceptions reported by `bugsnag-unity` V7+ will not group automatically in the Bugsnag dashboard with the equivalent exception reported by older versions of the library. 
+
+All errors will still be reported, but you may wish to pay extra attention to new exceptions that come through once upgrading and group them accordingly.
+
+Handled errors and native crashes will continue grouping in the same way and are not affected by this upgrade.
```
gareththackeray, line 12:
```suggestion
Handled errors and native crashes will maintain their existing groups and are not affected by this upgrade.
```

CHANGELOG.md
```diff
@@ -1,13 +1,23 @@
 # Changelog
 
-## 7.0.0 (2022-04-23)
+## 7.0.0 (2022-05-30)
 
 This version contains **breaking** changes: some errors generated by the new version will no longer group in the Bugnsag dashboard with equivalent errors generated by older versions. This is due to the introduction of a new exception interceptor that intercepts the C# exceptions objects, rather than parsing the information from Unity logs.
```
gareththackeray, line 5:
```suggestion
This version contains **breaking** changes: some errors generated by the new version will no longer group in the Bugnsag dashboard with equivalent errors generated by older versions. This is due to the introduction of a new exception interceptor that gets passed the C# exception objects, rather than having to parse the information from Unity logs.
```

gareththackeray (approved):

PR #561, opened 2022-05-27
Release v6.3.2

rich-bugsnag:
## Goal

Hotfix release to fix an issue where leaving a breadcrumb that contained metadata with a null value caused an exception.

## Changeset

Added a null check to android NativeInterface.BuildJavaMapDisposable(IDictionary<string, object> src)

## Testing

Added CI tests for leaving a breadcrumb with a null metadata value to all platforms


twometresteve (approved):
LGTM

PR #562, opened 2022-05-27
Increase timeout for macOS e2e tests

twometresteve:
## Goal

Stop e2e tests failing unnecessarily through timeouts.

## Design

These test normally take 25-30 minutes to run, so a timeout of 30 is just too restrictive.

## Testing

Covered by CI.


rich-bugsnag (approved):
Thanks!

PR #565, opened 2022-06-01
8524 move cache manager to interface

rich-bugsnag:
## Goal

Future platforms will require platform specific code to manage cached data on device storage. To support this, the existing cache storage code must be broken out into an interface.

## Design

The following interface has been implemented

public interface ICacheManager
{
    string GetCachedDeviceId();
    void SaveDeviceIdToCache(string id);
    void CacheSession(string id, string json);
    void CacheEvent(string id, string json);
    void RemoveCachedEvent(string id);
    void RemoveCachedSession(string id);
    string[] GetCachedPayloadIds();
    IPayload GetCachedPayload(string id);
}

## Changeset

- Created ICacheManager.
- Created CacheManager fallback implementation that is used by Fallback, Android, iOS, MacOS and Windows projects.
- Created PayloadManager class to handle pending payloads before writing to the cache.

## Testing

Manually tested and covered by existing CI tests for persistence on each platform


src/BugsnagUnity/ICacheManager.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using BugsnagUnity.Payload;
+
+namespace BugsnagUnity
+{
+    public interface ICacheManager
```
kattrali, line 6: Do you need something here to set the max number of cached events/sessions?

rich-bugsnag: the fallback doesnt need that as it has the config but i can see that perhaps the native plugin will need it.
but there are a few things, like setMaxCacheSize that also might be needed so i'm going to only implement what is currently needed in this PR and make changes as needed when the time comes later on 👍

src/BugsnagUnity/ICacheManager.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        string GetCachedDeviceId();
+        void SaveDeviceIdToCache(string id);
+        void CacheSession(string id, string json);
+        void CacheEvent(string id, string json);
```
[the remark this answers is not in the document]
rich-bugsnag: Done 👍

kattrali (approved):

PR #570, opened 2022-06-13
fix(ci): remove csharp layer cache inbetween scenarios

rich-bugsnag:
## Goal

- CSharp layer cache data was not being removed before each scenario 

## Changeset

- Added method to clear Csharp layer data between scenarios 

## Testing

Covered by existing tests


twometresteve (approved):
LGTM pending CI.

PR #571, opened 2022-06-14
fix(sessions): remove 1 frame wait before sending initial session

rich-bugsnag:
## Goal

Due to legacy code, events occuring after but during the same frame as Bugsnag.Start have no session data.

## Design

If autodetectsessions is enabled, the fallback will now create the initial session in the same frame as init.

## Changeset

- Removed 1 frame wait in fallback

## Testing

Added CI tests for iOS, Cocoa and Windows. MacOS does not work yet and will be worked on by the cocoa team at a later date


lemnik (approved):
LGTM

PR #572, opened 2022-06-16
Convert desktop tests to use Maze Runner Commands

twometresteve:
## Goal

Convert the desktop tests (macOS, Windows and webGL) to use Maze Runner commands.

## Design

Rather than setting environment variables, the test fixture will make and HTTP request to Maze Runner's `/command` endpoint to determine what it should do next.  The call is made from `Main.Start`.

The eventual aim (not in this PR) will be to combine the mobile tests to use the same mechanism and set of feature files.

## Changeset

- Add call to `/command` and associated processing.
- We now need to make running of the test fixture a non-blocking call.  As part of this I've added a Cucumber step to send a Command to explicitly quit the app, rather than having the app close itself on a timer.
- Remove code relating to setting and processing of environment variables.
- Some scenarios that were duplicated for macOS/webGL have now been combined.
- `PreventCrashPopups` added to the native bundle to try and prevent the app from displaying a popup after crashing multiple times (and breaking all subsequent scenarios).  
 
## Testing

Covered by a full CI run.


features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -42,37 +98,8 @@ public void Start()
...
-            Bugsnag.ClearMetadata("init");
-            Bugsnag.ClearMetadata("test", "test2");
-        }
-        RunScenario(scenario);
-    }
 
-    void CloseApplication()
-    {
-        Application.Quit();
+        StartCoroutine(RunNextMazeCommand());
```
twometresteve, line 102: Here...

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -7,114 +8,109 @@
...
 
-    private float _closeTime = 5;
+#if UNITY_STANDALONE || UNITY_WEBGL
+    private string _mazeHost = "http://localhost:9339";
+#else
+    private string _mazeHost = "http://bs-local.com:9339";
+#endif
 
-    public void Start()
+    IEnumerator RunNextMazeCommand()
```
rich-bugsnag, line 50: Nit, would be nice if Main.Start was the first method in the file to save future devs time working out entry points etc...

twometresteve: Moved.

rich-bugsnag (approved):
Nice!

PR #576, opened 2022-06-27
task(config): add telemetry config option to set value for android and cocoa

rich-bugsnag:
## Goal

add telemetry config option to set value for android and cocoa

## Changeset

- Added TelemetryType enum
- Added config.Telemetry 
- Pass value to Android and Cocoa on start

## Testing

Manually tested (with logging) that the value is set correctly in the Android and Cocoa SDKs

### DIFF
--- a/src/BugsnagUnity.m
+++ b/src/BugsnagUnity.m
@@ -546,7 +546,7 @@ void bugsnag_setEnabledTelemetryTypes(const void *configuration, const char *typ
         return;
     }
 
-    ((__bridge BugsnagConfiguration *)configuration).telemetry &= ~BSGTelemetryInternalErrors;
+    ((__bridge BugsnagConfiguration *)configuration).telemetry = 0;
     for (int i = 0; i < count; i++) {
         const char *enabledType = types[i];
         if (enabledType != nil) {
--- a/src/BugsnagUnity.m
+++ b/src/BugsnagUnity.m
@@ -19,6 +19,18 @@
         const char *user_email;
     };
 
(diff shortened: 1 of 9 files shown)


src/BugsnagUnity.m
```diff
@@ -528,6 +540,25 @@ void bugsnag_setEnabledBreadcrumbTypes(const void *configuration, const char *ty
...
+    }
+
+    ((__bridge BugsnagConfiguration *)configuration).telemetry &= ~BSGTelemetryInternalErrors;
+    for (int i = 0; i < count; i++) {
+        const char *enabledType = types[i];
+        if (enabledType != nil) {
+            NSString *typeString = [[NSString alloc] initWithUTF8String:enabledType];
+            if([typeString isEqualToString:@"InternalErrors"])
+            {
+                ((__bridge BugsnagConfiguration *)configuration).telemetry |= BSGTelemetryInternalErrors;
```
nickdowell, line 556: This differs from the Android behaviour, where unknown telemetry options will be disabled.

Maybe make it match like so?

```suggestion
    ((__bridge BugsnagConfiguration *)configuration).telemetry = 0;
    for (int i = 0; i < count; i++) {
        const char *enabledType = types[i];
        if (enabledType != nil) {
            NSString *typeString = [[NSString alloc] initWithUTF8String:enabledType];
            if([typeString isEqualToString:@"InternalErrors"])
            {
                ((__bridge BugsnagConfiguration *)configuration).telemetry |= BSGTelemetryInternalErrors;
```

nickdowell (approved):
Cocoa side looks ok 👍

PR #591, opened 2022-07-20
Update bugsnag-cocoa to v6.21.0

nickdowell:
`bsg_ksmachfreeMemory()` has been removed from bugsnag-cocoa, so `BugsnagUnity.m` has been updated accordingly.


rich-bugsnag (approved):
Thanks!

PR #592, opened 2022-07-20
refac(client): simplify fallback usage detection

rich-bugsnag:
## Goal

Fallback detection was over complicated and stopped some things working on a new native platform, so i simplified it. 

## Testing

Covered by existing CI tests


twometresteve (approved):
LGTM

PR #594, opened 2022-08-03
Test flake: Explicitly set offset from UTC when creating date/times

twometresteve:
## Goal

Fix tests that fail when run on a device whose timezone is anything but GMT.

## Changeset

Explicitly set the time using an offset to prevent misinterpretation occurring during casting.

## Testing

Ran the before and after using an Appium capability to set the device timezone to New_York - failed before and passed afterards.

### DIFF
--- a/features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
@@ -221,7 +221,7 @@ public class MobileScenarioRunner : MonoBehaviour {
                     device.Jailbroken = true;
                     device.CpuAbi = new string[] { "poo", "baar" };
                     device.Orientation = "Custom Orientation";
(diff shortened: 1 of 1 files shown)


features/fixtures/maze_runner/Assets/Scripts/MobileScenarioRunner.cs
```diff
@@ -221,7 +221,7 @@ private Configuration PreapareConfigForScenario(string scenarioName)
                     device.Jailbroken = true;
                     device.CpuAbi = new string[] { "poo", "baar" };
                     device.Orientation = "Custom Orientation";
-                    device.Time = new DateTime(1985, 08, 21, 01, 01, 01);
+                    device.Time = new DateTimeOffset(1985, 08, 21, 01, 01, 01, new TimeSpan(0));
```
gareththackeray, line 224: Slightly more idiomatic but same result
```suggestion
                    device.Time = new DateTimeOffset(1985, 08, 21, 01, 01, 01, TimeSpan.Zero);
```

gareththackeray (approved):

PR #596, opened 2022-08-09
Build an nspd

twometresteve:
## Goal

Changes the build mechanism for Switch so that it creates a `.nspd` folder, which can then be converted into an installable `.nsp` file using the Switch AuthoringTool.

## Changeset

Build method updated to follow the pattern that iOS/Android use.

## Testing

I have a branch now ready for review in the Switch repo, which successfully builds the nsp.  A basic CI run is sufficient here to show that it does not break bugsnag-unity.


joshedney (approved):
LGTM!

PR #599, opened 2022-08-17
Reinstate Windows build of plugin (dev purposes only)

twometresteve:
## Goal

Reinstate building of the plugin on Windows on CI.   

## Design

This is just for development purposes, to ensure that we don't break it for ourselves during development.  We do not ship the built package.  This step was broken when we moved to using WSL on CI, but fixed itself again after updating the build servers fro 2019 to 2022.

## Changeset

Skipping of Buildkite step removed, resulting artifact uploaded for some extra confidence that it's built properly (under a different name to avoid conflicting with the real one).

## Testing

Covered by a basic CI run.


rich-bugsnag (approved):
Cool, thanks!

PR #602, opened 2022-08-23
e2e testing on Nintendo Switch

twometresteve:
## Goal

Get e2e tests running on Nintendo Switch.

@joshedney I would be grateful if you could sanity check this PR in conjunction with the associate PR for Switch (private repo) before I put it into full reivew.

## Design

This change extends the existing use of the Maze Runner `/command` pattern to control the behaviour of the Switch during tests.

Building the test fixture and running the tests is documented in our private Switch repo.

## Changeset

Additional logic added as needed for Switch, utilising Nintendo SDK executables as needed to launch/terminate the test fixture app.

## Testing

All e2e tests now run against our Switch dev kit (with some failures to be processed).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'xcodeproj'
 
 unless Gem.win_platform?
   # Use official Maze Runner release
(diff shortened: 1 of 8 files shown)


rich-bugsnag (approved):
Looks Good 👍

PR #604, opened 2022-08-25
bugFix(editor): set don't save hide flags on generated gameObjects

rich-bugsnag:
## Goal

Users have reported gameobjects generated by us persisting in the editor when they shouldn't.

I was unable to reproduce the particular issue reported, but this fix is harmless and actually a good idea regardless as it adds extra protection should this bug be a real issue in future versions of unity.

## Changeset

- Simplified the creation of the MainThreadDispatchHandler class.
- Added DontSave flags to the MainThreadDispatchHandler and the TimingTracker gameObjects.

## Testing

Covered by existing tests.

### DIFF
--- a/src/BugsnagUnity/TimingTrackerBehaviour.cs
+++ b/src/BugsnagUnity/TimingTrackerBehaviour.cs
@@ -14,6 +14,7 @@ namespace BugsnagUnity
         {
             // Make sure that the tracker persists accross scenes.
             DontDestroyOnLoad(gameObject);
+            gameObject.hideFlags = HideFlags.DontSave;
         }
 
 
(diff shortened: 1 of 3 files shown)


src/BugsnagUnity/TimingTrackerBehaviour.cs
```diff
@@ -14,6 +14,7 @@ private void Awake()
         {
             // Make sure that the tracker persists accross scenes.
             DontDestroyOnLoad(gameObject);
+            gameObject.hideFlags = HideFlags.DontSave;
```
nickdowell, line 17: Is it better to use [`HideFlags.DontSave`](https://docs.unity3d.com/ScriptReference/HideFlags.DontSave.html) or [`HideFlags.HideAndDontSave`](https://docs.unity3d.com/ScriptReference/HideFlags.HideAndDontSave.html) ?

rich-bugsnag: I don't want to hide the object because it makes troubleshooting potential future issues much harder both for users and for us.

nickdowell (approved):
LGTM

PR #605, opened 2022-08-26
refac(android): simplify java string creation [full ci]

rich-bugsnag:
## Goal

A change (possibly a bug) in the Unity code base caused existing notifier code in the android NativeInterface to crash.

This prompted us to examin the code and we decided that it could be simplified to resolve the issue and hopfully future proof it.

## Changeset

- Removed the `BuildJavaStringDisposable` method and replaced calls to it with simple usage of the `new AndroidJavaObject("java.lang.String", cSharpString);` style java string constructor.

## Testing

Covered by existing CI tests and manually tested on android 6 and 12 devices with a string containing a null specifier like so "test\0test".


lemnik (approved):
LGTM

PR #610, opened 2022-09-05
fix(stacktraces): fix for missing frames in exceptions occurring in async methods

rich-bugsnag:
## Goal

The new exception handler was missing stacktrace frames from exceptions that happened in async methods. 

This method `System.Diagnostics.StackTrace(exception, true).GetFrames();` Was ignoring any frame above the line `--- End of stack trace from previous location where exception was thrown ---`.

I changed it to use our own stacktrace parser and there seems to be no difference in the formatting so it should be a good fix.

## Changeset

- Changed the new exception handler to use the same method of trace generation as is used when generating a trace from unity log events.

## Testing

Manually tested grouping and added new async CI tests.

### DIFF
--- a/src/BugsnagUnity/Payload/Error.cs
+++ b/src/BugsnagUnity/Payload/Error.cs
@@ -133,10 +133,10 @@ namespace BugsnagUnity.Payload
 
         public string Type { get => "Unity"; }
 
+
         internal static Error FromSystemException(System.Exception exception, System.Diagnostics.StackFrame[] alternativeStackTrace)
-        {
+        {            
             var errorClass = exception.GetType().Name;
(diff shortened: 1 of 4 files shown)


src/BugsnagUnity/Payload/Error.cs
```diff
@@ -133,10 +133,10 @@ public string ErrorMessage
 
         public string Type { get => "Unity"; }
 
+
         internal static Error FromSystemException(System.Exception exception, System.Diagnostics.StackFrame[] alternativeStackTrace)
-        {
+        {
```
nickdowell, line 138:
```suggestion
        {
```

tomlongridge (approved):
LGTM - needs a changelog entry though. Happy if you want to do that in the impending release.

PR #611, opened 2022-09-05
Remove Switch Application.nmeta file and ignore

twometresteve:
## Goal

Remove the Switch test fixture Application nmeta file.

## Design

The file included a path to an application icon file that in general will not exist on another user's machine (and does not on our CI servers).  My proposal is that we add the file to our Switch-specific repo and copy it into place for builds in the rakefile.  This would avoid PR churn in itself when it comes to making any changes for the Switch test fixture.

## Changeset

Application.nmeta removed and ignored (will be re-added in the Switch repo).

## Testing

Tested locally that the 'copy-in' approach will work.


rich-bugsnag (approved):
Thanks

PR #613, opened 2022-09-08
[PLAT-8881] Add Event.FeatureFlags

nickdowell:
## Goal

Allow event feature flags to be inspected in `OnSendError` callbacks.

## Changeset

Adds `FeatureFlags` getter to IEvent and implementations.

## Testing

Amends existing E2E scenarios to verify that `FeatureFlags` getter returns the expected values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* Added `BugsnagEvent.FeatureFlags` to allow feature flags to be queried before event delivery.
+  [#613](https://github.com/bugsnag/bugsnag-unity/pull/613)
+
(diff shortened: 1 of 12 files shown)


src/BugsnagUnity/Delivery.cs
```diff
@@ -57,9 +57,10 @@ public void Deliver(IPayload payload)
                                     return;
                                 }
                             }
-                            catch
+                            catch (Exception ex)
                             {
                                 // If the callback causes an exception, ignore it and execute the next one
+                                Debug.LogError("Ignoring exception thrown by OnSend callback: " + ex.ToString());
```
rich-bugsnag, line 63: This could end up in the users dashboard, do we want that?

rich-bugsnag (approved):
LGTM

PR #617, opened 2022-09-12
revert hide flags addition

rich-bugsnag:
## Goal

A bug fix introduced [here](https://github.com/bugsnag/bugsnag-unity/pull/604) caused null reff errors when using Bugsnag in the editor. 

This is due to that fact that when dont save hide flags are set then the objects must be destroyed manually, which, due the the nature of the SDK, is not possible without some major changes.

Because the original issue of gameobjects remaining after playmode is exited was not reproducable, we will revert the fix and investigate further before making any related changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 * Added `BugsnagEvent.FeatureFlags` to allow feature flags to be queried before event delivery.
   [#613](https://github.com/bugsnag/bugsnag-unity/pull/613)
 
+### Bug fixes
+
+* Reverted the hide flags bug fix introduced [here](https://github.com/bugsnag/bugsnag-unity/pull/604) as it was causing null ref errors when using Bugsnag in the editor. Please see this PR for more details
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
 * Added `BugsnagEvent.FeatureFlags` to allow feature flags to be queried before event delivery.
   [#613](https://github.com/bugsnag/bugsnag-unity/pull/613)
 
+### Bug fixes
+
+* Reverted the hide flags bug fix introduced [here](https://github.com/bugsnag/bugsnag-unity/pull/604) as it was causing null ref errors when using Bugsnag in the editor. Please see this PR for more details
```
tomlongridge, line 12:
```suggestion
* Reverted the `HideFlags` fix introduced [here](https://github.com/bugsnag/bugsnag-unity/pull/604) as it was causing (harmless) `NullReferenceException`s in the editor.
```

PR #619, opened 2022-09-13
[PLAT-8879] Store feature flags in OrderedDictionary

nickdowell:
## Goal

Improve performance of `AddFeatureFlag` APIs in apps that have large numbers of flags.

## Changeset

Switches to an [OrderedDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary) to store feature flags as names and variants.

## Testing

Existing E2E tests verify that storage works as expected.

Manually compared performance and found that adding 1000 feature flags has gone from 40ms to < 1ms.


rich-bugsnag (approved):
Very cool!

PR #620, opened 2022-09-13
Enable passing arguments when running Switch e2e tests

rich-bugsnag:
## Goal

Allow passing arguments to the test fixture when running Se2e tests on Switch.

These arguments will:
- (--mazeIP) provide the address of the Maze Runner server 
- (others) configure the native cache plugin

## Design

The `--mazeIp` argument must be given and is done so by setting the  `SWITCH_MAZE_IP` environment variable.  If the other args are left out then the fixture will operate with the default cache settings.

## Changeset

- Added getArgs plugin build to maze_runner
- Added code to apply those values to the fixture config before running scenarios
- Changed use of ControlTarget to RunOnTarget to allow use of arguments

## Testing

All desktop test successfully run on our Switch SDEV, others are covered by CI.


features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -34,23 +34,37 @@ public class Main : MonoBehaviour
     [DllImport("NativeCrashy")]
     private static extern void crashy_signal_runner(float num);
 
+#endif
```
twometresteve, line 37: Looks like the additions to this file are causing failures in the macOS/Windows tests due to differing line numbers.

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -34,23 +34,37 @@ public class Main : MonoBehaviour
...
+    private int _switchCacheIndex = 0;
+    private string _switchCacheName = string.Empty;
 
     public void Start()
     {
         Debug.Log("Maze Runner app started");
-
+#if UNITY_SWITCH
+    GetSwitchArguments();
+#endif
```
twometresteve, line 64: Would have been cleaner leaving it as the `#elif` on L52 (as was) and just changing L53 imo.

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -34,23 +34,37 @@ public class Main : MonoBehaviour
...
 #endif
 
     private const string API_KEY = "a35a2a72bd230ac0aa0f52715bbdc6aa";
     private Dictionary<string, string> _webGlArguments;
 
     private string _fakeTrace = "Main.CUSTOM () (at Assets/Scripts/Main.cs:123)\nMain.CUSTOM () (at Assets/Scripts/Main.cs:123)";
     private string _mazeHost;
+    private SwitchCacheType _switchCacheType = SwitchCacheType.R;
+    private int _switchCacheIndex = 0;
+    private string _switchCacheName = string.Empty;
```
twometresteve, line 57: Perhaps these should sit in the `#if` on L39?

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -67,6 +81,59 @@ public void Start()
...
+                    case "r":
+                        _switchCacheType = SwitchCacheType.R;
+                        break;
+                    case "i":
+                        _switchCacheType = SwitchCacheType.I;
+                        break;
+                    case "n":
+                        _switchCacheType = SwitchCacheType.None;
+                        break;
+                }
```
twometresteve, line 119: I wouldn't want to go overboard on input validation when we're in control of it, but a `default:` here that says the argument is invalid and bails out could avoid some unexpected behaviour and lost time.

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -67,6 +81,59 @@ public void Start()
...
+                Debug.Log("Switch Cache Type set to: " + _switchCacheType);
+            }
+            if (arg == "--cacheIndex")
+            {
+                 _switchCacheIndex = int.Parse( bugsnag_getArg(i + 1));
+                Debug.Log("Switch cache index set to: " + _switchCacheIndex);
+            }
+            if (arg == "--mountName")
```
twometresteve, line 127: Any reason not to call this `cacheName` for consistency and memorability?

twometresteve: Ah, I see the underlying config value is SwitchCacheMountName.  I think I'll call it `--cacheMountName`.  Longer, but more consistent imo.

features/fixtures/maze_runner/Assets/Scripts/Main.cs
twometresteve: Added

features/fixtures/maze_runner/Assets/Scripts/Main.cs
twometresteve: Moved

features/fixtures/maze_runner/Assets/Scripts/Main.cs
twometresteve: Moved back

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -67,6 +81,59 @@ public void Start()
...
+
+        for (int i = 0; i < count; i++)
+        {
+            var arg = bugsnag_getArg(i);
+            if (!arg.Contains("--"))
+            {
+                continue;
+            }
+            Debug.Log("CHECKING ARG: " + arg);
+            if (arg == "--mazeIp")
```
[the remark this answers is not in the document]
twometresteve: Changed

features/fixtures/maze_runner/Assets/Scripts/Main.cs
twometresteve: Fixed

features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -36,6 +36,19 @@ public class Main : MonoBehaviour
...
+
+    [DllImport("__Internal")]
+    private static extern int bugsnag_getArgsCount();
+
+    [DllImport("__Internal")]
+    private static extern string bugsnag_getArg(int index);
+
+    private SwitchCacheType _switchCacheType = SwitchCacheType.R;
+    private int _switchCacheIndex = 0;
+    private string _switchCacheMountName = string.Empty;
```
rich-bugsnag, line 49:
```suggestion
    private string _switchCacheMountName = "BugsnagCache";
```
This is probably why that mount name error happened

twometresteve: Ah, so a value is needed and the "else" value I put on L369 of unity_steps.rb would not be needed if we change this.

twometresteve: I've changed both

twometresteve (approved):
I'm happy with your changes and we're discussed IRL that you're happy with mine.

PR #624, opened 2022-09-16
fix(CI): remove getArgs lib from fixture

rich-bugsnag:
## Goal

remove getArgs lib from fixture as it was causing build issues for android in unity 2018.

The lib will be copied over when needed on other platforms.


twometresteve (approved):
LGTM pending CI

PR #625, opened 2022-09-20
Allow running of indexed cache tests on Switch

twometresteve:
## Goal

Provide the ability to run the same e2e tests that run on regular cache test fixture, on the indexed cache test fixture.

## Changeset

Moves the existing logic for the reading of `SWITCH_MAZE_IP` to the Cucumber `BeforeAll` hook and adds a check on `SWITCH_CACHE_TYPE` (if it is set, it must be meaningful).

## Testing

Tested locally that the general mechanism works with our dev kit.

### DIFF
--- a/features/steps/unity_steps.rb
+++ b/features/steps/unity_steps.rb
@@ -356,17 +356,14 @@ end
 
 def switch_run_on_target
   # Maze IP must always be provided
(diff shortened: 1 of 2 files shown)


rich-bugsnag (approved):
LGTM 👍

PR #626, opened 2022-09-20
Switch configs in settings UI

rich-bugsnag:
## Goal

add switch specific config values to the config UI

Screenshot 2022-09-20 at 11 33 20

### DIFF
--- a/src/Assets/Bugsnag/Editor/BugsnagEditor.cs
+++ b/src/Assets/Bugsnag/Editor/BugsnagEditor.cs
@@ -15,7 +15,7 @@ namespace BugsnagUnity.Editor
 
         private bool _showBasicConfig = true;
 
-        private bool _showAdvancedSettings, _showAppInformation, _showEndpoints, _showEnabledErrorTypes;
+        private bool _showAdvancedSettings, _showAppInformation, _showEndpoints, _showEnabledErrorTypes, _showSwitch;
 
         public Texture DarkIcon, LightIcon;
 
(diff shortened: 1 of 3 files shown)


src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
```diff
@@ -40,6 +40,12 @@ public class BugsnagSettingsObject : ScriptableObject
         public double SecondsPerUniqueLog = 5;
         public List<TelemetryType> Telemetry = new List<TelemetryType> { TelemetryType.InternalErrors };
         public int VersionCode;
+
+        public SwitchCacheType SwitchCacheType = SwitchCacheType.R;
+        public string SwitchCacheMountName = "BugsnagCache";
+        public int SwitchCacheIndex = 0;
+        public int SwitchMaxCacheSize = 10485760;
```
tomlongridge, line 47: Suggest we rename this for consistency - perhaps that requires the underlying config option to change too? But feels worth it to save confusion.

```suggestion
        public int SwitchCacheMaxSize = 10485760;
```

src/Assets/Bugsnag/Editor/BugsnagEditor.cs
```diff
@@ -105,6 +105,14 @@ private void DrawSettingsEditorWindow()
             {
                 DrawEndpoints(so);
             }
+
+            GUILayout.Space(5);
+            _showSwitch = EditorGUILayout.Foldout(_showSwitch, "Nintendo Switch (Requires Nintendo Switch Bugsnag plugin)", true);
```
gareththackeray, line 110: I imagined this in a tooltip is that not possible on a group header?

rich-bugsnag: Done 👍

src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
rich-bugsnag: Done

gareththackeray (approved):

PR #636, opened 2022-10-06
refak(CI): convert the unity fixture to use scenario classes. part 1

rich-bugsnag:
## Goal

This is the first part of a multipart refactor of the way CI runs in the Unity repo.

The end result will be all tests running with scenario classes and all platforms running using mazerunner commands.

## Design

There will be a csharp features folder that contains tests that should run on all platforms.

Each platform will have it's own feature files to test platform specific stuff, similar to how it is now but better organised.

## Changeset

- Added scenario classes and code to find/run these classes

## Testing

The new tests files will run on windows and MacOS for now, later the mobile fixture will be converted to use maze commands and then run the new features.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -22,24 +22,24 @@ steps:
         - exit_status: "*"
           limit: 1
 
-  - label: Ensure notifier builds on Windows (for development)
-    timeout_in_minutes: 30
-    agents:
-      queue: windows-unity-wsl
-    env:
-      UNITY_VERSION: "2018.4.36f1"
-      WSLENV: UNITY_VERSION
-    command:
-      - /mnt/c/Windows/System32/cmd.exe /c  .\\scripts\\ci-build-windows-plugin.bat
-    plugins:
-      artifacts#v1.5.0:
-        upload:
-          - from: Bugsnag.unitypackage
-            to: Bugsnag_WindowsBuilt.unitypackage
-    retry:
-      automatic:
-        - exit_status: "*"
-          limit: 1
+  # - label: Ensure notifier builds on Windows (for development)
+  #   timeout_in_minutes: 30
+  #   agents:
+  #     queue: windows-unity-wsl
+  #   env:
+  #     UNITY_VERSION: "2018.4.36f1"
+  #     WSLENV: UNITY_VERSION
+  #   command:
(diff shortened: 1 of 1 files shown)


features/fixtures/maze_runner/Assets/Scripts/Scenarios/Fallback/Breadcrumbs/DisableBreadcrumbs.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public override void PrepareConfig(string apiKey, string host)
+    {
+        base.PrepareConfig(apiKey, host);
+        Configuration.EnabledBreadcrumbTypes = new BreadcrumbType[0];
+    }
+
+    public override void Run()
+    {
+        base.Run();
```
nickdowell, line 18: I think you can remove all these calls to `base.Run()`, they just add noise.

features/fixtures/maze_runner/Assets/Scripts/Scenarios/Fallback/Breadcrumbs/EnableBreadcrumbs.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using BugsnagUnity;
+using System;
```
nickdowell, line 5: I suspect we don't need all these `using` statements

features/fixtures/maze_runner/Assets/Scripts/Scenarios/Fallback/Breadcrumbs/DisableBreadcrumbs.cs
rich-bugsnag: done 👍

nickdowell (approved):
LGTM!

Gemfile
```diff
@@ -6,7 +6,7 @@ gem 'xcodeproj'
 
 unless Gem.win_platform?
   # Use official Maze Runner release
-  gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.19.1'
+  gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v7.2.1'
```
twometresteve, line 9: Upgrading to v7 at the same time is probably too much to take on in one PR - note that @Cawllec is already working on that.

PR #641, opened 2022-10-10
refac(CI): add persistence scenario classes

rich-bugsnag:
## Goal

add persistence scenario classes

## Testing

Same coverage as before

### Note: 
Session persistence and webGL tests are not working and will be investigated at a later date as this is not a change in current coverage.


Cawllec (approved):
LGTM, providing it passes on CI.  I'm trusting that the moved scenario files that haven't been modified are still desired and working :)

PR #646, opened 2022-10-13
task (CI): create metadata scenario classes

rich-bugsnag:
## Goal

create metadata scenario classes

## Testing

Existing coverage replaced and added more coverage


twometresteve:
Maybe a `MetadataHelper` class could be used to save repeating the same setup code in three places?  Looks fine apart from that though.

twometresteve (approved):
I see now, not really commonizable in practice.

Cawllec (approved):

PR #648, opened 2022-10-14
fix(UWP): skip device memory and space metadata on UWP builds

rich-bugsnag:
## Goal

Remove `device.freeDisk` and `device.freeMemory` metadata from events sent in [UWP](https://docs.unity3d.com/Manual/windowsstore-buildsettings.html) builds as accessing the data causes an exception.

## Design

This is a temporary fix, we plan to implement a safe way to access this data in the future.

## Changeset

- Added a check to only access said data in windows standalone builds

## Testing

Covered by existing tests on standalone builds. We plan to add UWP builds to CI in the future.


nickdowell (approved):
LGTM, assuming this approach has been agreed 👍

PR #653, opened 2022-10-25
fix(cache): hide any potential File IO exceptions

rich-bugsnag:
## Goal

There have been reports of the csharp cache manager throwing exceptions on Huawei devices running Android 10.

File IO apis can rarely be 100% safe, so I have added try catch blocks around access to the File class so that any exceptions raised within the cache manager are silenced.

## Testing

Covered by existing tests

### DIFF
--- a/src/BugsnagUnity/Native/Fallback/CacheManager.cs
+++ b/src/BugsnagUnity/Native/Fallback/CacheManager.cs
@@ -73,7 +73,11 @@ namespace BugsnagUnity
 
         public void SaveDeviceIdToCache(string deviceId)
         {
-            File.WriteAllText(_deviceIdFile, deviceId);
+            try
+            {
+                File.WriteAllText(_deviceIdFile, deviceId);
+            }
+            catch { }
         }
 
         private void RemoveExpiredPayloads()
@@ -111,7 +115,11 @@ namespace BugsnagUnity
 
         private void WritePayloadToDisk(string jsonData, string path)
         {
-            File.WriteAllText(path, jsonData);
+            try
+            {
+                File.WriteAllText(path, jsonData);
+            }
+            catch{ }
         }
 
         private void CheckForMaxCachedPayloads(string[] payloads, int maxPayloads)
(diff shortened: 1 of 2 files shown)


src/BugsnagUnity/Native/Fallback/CacheManager.cs
```diff
@@ -124,41 +132,56 @@ private void CheckForMaxCachedPayloads(string[] payloads, int maxPayloads)
...
         {
-            var ordered = filePaths.OrderBy(file => File.GetCreationTimeUtc(file)).ToArray();
-            foreach (var file in ordered.Take(numToRemove))
+            try
             {
-                File.Delete(file);
+                var ordered = filePaths.OrderBy(file => File.GetCreationTimeUtc(file)).ToArray();
+                foreach (var file in ordered.Take(numToRemove))
+                {
+                    File.Delete(file);
```
twometresteve, line 140: If one delete fails then it will bail out of the loop and not try any others - is that what you want?  I might have thought that wrapping the try/catch only around the `Delete` line itself would be better?  (Same for the other methods of course).

rich-bugsnag: fixed 👍

twometresteve: And the other methods in the file?

src/BugsnagUnity/Native/Fallback/CacheManager.cs
```diff
@@ -44,7 +44,14 @@ public CacheManager(Configuration configuration)
...
             CheckForDirectoryCreation();
-            RemoveExpiredPayloads();
+            try
+            {
+                RemoveExpiredPayloads();
+            }
+            catch
+            {
+                //not avaliable in unit tests
+            }
```
nickdowell, line 54: Since `RemoveExpiredPayloads()` catches any exceptions itself, do we need a `try` block here?

rich-bugsnag: yes, because even accessing _cachedEvents during a unit test causes the test to fail

rich-bugsnag: i should change the way we access _cachedEvents and _cachedSessions to make this cleaner.

rich-bugsnag: Done 👍

nickdowell (approved):

PR #656, opened 2022-10-25
refac(cache): make session and event removal code more readable

rich-bugsnag:
## Goal

make session and event removal code more readable

## Testing

Covered by existing CI tests

### DIFF
--- a/src/BugsnagUnity/Native/Fallback/CacheManager.cs
+++ b/src/BugsnagUnity/Native/Fallback/CacheManager.cs
@@ -166,32 +166,32 @@ namespace BugsnagUnity
 
         public void RemoveCachedEvent(string id)
         {
-            try
-            {
-                foreach (var cachedEventPath in GetCachedEventFiles())
-                {
-                    if (cachedEventPath.Contains(id))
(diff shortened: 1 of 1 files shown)


src/BugsnagUnity/Native/Fallback/CacheManager.cs
```diff
@@ -166,32 +166,32 @@ private void RemoveOldestFiles(string[] filePaths, int numToRemove)
...
-                    }
+                    DeleteFile(path);
+                    return;
                 }
             }
-            catch { }
         }
 
-        public void RemoveCachedSession(string id)
+        private void DeleteFile(string path)
```
twometresteve, line 189: I see the idea behind this method, but it's only called from one place currently.  Did you mean to move the other instances of `File.Delete` across to use it (and remove their try/catch blocks)?

rich-bugsnag: done and added a similar WriteFile method to reduce try catch blocks 👍

twometresteve (approved):

PR #659, opened 2022-10-27
task(CI): add switch scenario that crashes the application

rich-bugsnag:
## Goal

add switch scenario that crashes the application


twometresteve:
Useful, but just to note that it's not for merging.

PR #660, opened 2022-10-27
Bump timeouts and OS versions

twometresteve:
## Goal

Bump some step timeouts and OS versions (in the hope that they run more reliably/quickly).

## Testing

Covered by a full CI run.


rich-bugsnag (approved):
This fixed iOS, thanks 👍

PR #661, opened 2022-10-27
task(CI): create mobile specific scenario classes

rich-bugsnag:
## Goal

create mobile specific scenario classes

## Changeset

Added tests for the following 

### Note: This may not seem like many tests compared to what used to run in the Android and ios feature folders, but that is because 90% of legacy tests have been consolodated into the csharp feature folder.

For example, before, each platform had a seperate test for csharp layer unhandled exceptions, now each platform shares the test in the csharp/events.feature file.

Android
- On Send Native Callback
- Android Persistence Directory
- Android JVM Smoke Test
- Android JVM Background Thread Smoke Test
- Android Last Run Info
- Android Disable Crashes
- Android NDK Signal 

iOS
- On Send Native Callback
- Disable Crashes
- Last Run Info
- iOS native exception Smoke Test
- iOS signal Smoke Test

Switch
- SwitchCacheType set to None
- Max Cache Size
- Switch Metadata

## native config tests

Some native only config value tests have been ported over, however the tests for Configuration.SendThreads have not been moved over as native config testing will soon be addressed in PLAT-9131. 

Testing the functionality of those values is a duplication of tests accross platforms and takes up testing run time. The raised task is to implement a test that confirms that platform specific config values have been correctly passed through to the native layers, but will not spend time validating their behaviours as that is (should be) covered by the native notifiers themselves.


features/ios/ios_native_errors.feature
```diff
@@ -1,65 +1,114 @@
...
-        And I relaunch the Unity mobile app
-        When I run the "Check Last Run Info" mobile scenario
-        Then I wait to receive 2 errors
-        And I discard the oldest error
-        And the exception "message" equals "Last Run Info Correct"
\ No newline at end of file
+        And I clear any error dialogue
+        And On Mobile I relaunch the app
+        And I run the game in the "StartSDKDefault" state
+        And I wait for 3 seconds
```
twometresteve, line 74: No need to wait a few seconds before the next step, as it waits anyway.

rich-bugsnag: Done 👍

twometresteve (approved):

PR #662, opened 2022-10-30
Refactor of all CI functionality

rich-bugsnag:
## Goal

This integration branch contains a complete rewrite of the way that CI tests function.

The main goals are:

- Replace current  test running code with unified scenario classes.
- Run all tests on all platforms with Maze commands instead of CL arguments or UI input events.
- Reduce duplicated tests and run unified Csharp layer tests on all platforms.

### Notifier Bugs
During development, several bugs were identified and have been ticketed and will be fixed in the near future.

### Flakes
Due to the large device and Unity version converage, flakes still occur from time to time. The good news is that now they are much easier to investigate and act on.


twometresteve (approved):
LGTM - let's merge!  

I have raised #664 to clean up a few things, but that can go into `next` after getting this in.

PR #666, opened 2022-11-07
PLAT-8717 Add usage telemetry option

rich-bugsnag:
## Goal

Add usage telemetry option and pass it through to Android and Cocoa

## Testing

Added smoke test


features/android/android_config.feature
```diff
@@ -7,3 +7,15 @@ Feature: Android Config
     When I run the game in the "AndroidPersistenceDirectory" state
     And I wait to receive an error
     And the exception "message" equals "Directory Found"
+
+  Scenario: Telemetry
+    When I run the game in the "AndroidBackgroundJVMSmokeTest" state
+    And I wait for 2 seconds
+    And I clear any error dialogue
```
twometresteve, line 14: Is this really needed?

features/ios/ios_config.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Scenario: Telemetry
+    When I run the game in the "IosNativeException" state
+    And I wait for 2 seconds
+    And I clear any error dialogue
+    And On Mobile I relaunch the app
+    And I run the game in the "StartSDKDefault" state
+    And I wait to receive an error
+    And the error payload field "events.0.usage.config.maxBreadcrumbs" equals 50
+    And the error payload field "events.0.usage.callbacks.onSession" equals 1
```
twometresteve, line 14: I'm missing the purpose of these new scenarios, which don't introduce new scenario classes.  Do they somehow demonstrate that the new config works, or is there not actually any underlying functionality in place at the moment - so we're just verifying that it doesn't crash?  (If the latter, are existing test not sufficient for that?).

features/android/android_config.feature
rich-bugsnag: The usage telemetry data will only be present in android and cocoa events, hence a seperate scenario needed to check it exists

features/ios/ios_config.feature
rich-bugsnag: The usage telemetry data will only be present in android and cocoa events, hence a seperate scenario needed to check it exists

features/android/android_config.feature
twometresteve: Here I meant the specific step `I clear any dialogue`

features/ios/ios_config.feature
twometresteve: Sure, but there's already scenarios that seem to do the same thing ("iOS native exception Smoke Test" and "Android JVM Background Thread Smoke Test".

Can't you just add the two "the error payload field blah is" steps to those existing scenarios, saving the additional runtime and potential network flakes?

features/android/android_config.feature
rich-bugsnag: Yes it might be, sometimes the OS shows the "this app keeps crashing" popup

features/ios/ios_config.feature
rich-bugsnag: done

twometresteve (commented):
Test bits look good to me, not so well qualified to judge on the notifier changes.

tomlongridge (approved):

PR #671, opened 2022-11-18
PLAT-8627 truncate breadcrumbs when payload oversized

rich-bugsnag:
## Goal

truncate breadcrumbs when payload oversized

## Design

If. a payload is bigger than 1MB and truncating the strings in metadata does not bring it below 1MB, then breadcrumbs are removed until there are none left or the payload is under 1MB

## Changes

- Raised max breadcrumbs default to 100
- Set max value of max breadcrumbs to 500
- Added check for breadcrumb truncation to the Delivery class

## Testing

Added E2E test

### DIFF
--- a/src/BugsnagUnity/Delivery.cs
+++ b/src/BugsnagUnity/Delivery.cs
@@ -33,6 +33,19 @@ namespace BugsnagUnity
 
         private const string STRING_TRUNCATION_MESSAGE = "***{0} CHARS TRUNCATED***";
 
+        private const string BREADCRUMB_TRUNCATION_MESSAGE = "Removed, along with {0} older breadcrumbs, to reduce payload size";
+
+        private const string EVENT_KEY = "event";
+
+        private const string BREADCRUMBS_KEY = "breadcrumbs";
+
+        private const string METADATA_KEY = "metaData";
+
+        private const string TYPE_KEY = "type";
+
+        private const string NAME_KEY = "name";
+
+
 
         internal Delivery(Client client, Configuration configuration, CacheManager cacheManager, PayloadManager payloadManager)
         {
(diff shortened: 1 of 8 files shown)


src/BugsnagUnity/Delivery.cs
```diff
@@ -33,6 +33,19 @@ class Delivery
...
+
+        private const string EVENT_KEY = "event";
+
+        private const string BREADCRUMBS_KEY = "breadcrumbs";
+
+        private const string METADATA_KEY = "metaData";
+
+        private const string TYPE_KEY = "type";
+
+        private const string NAME_KEY = "name";
```
tomlongridge, line 46: Just querying if there's any better home for these constants?

I would also suggest naming them a little more clearly so they say what they do rather than describe the content of the string:

```suggestion
        private const string EVENT_KEY_EVENT = "event";

        private const string EVENT_KEY_BREADCRUMBS = "breadcrumbs";

        private const string EVENT_KEY_METADATA = "metaData";

        private const string EVENT_KEY_BREADCRUMB_TYPE = "type";

        private const string EVENT_KEY_BREADCRUMB_MESSAGE = "name";
        
        private const string EVENT_KEY_BREADCRUMB_METADATA = "metaData";
```

tomlongridge:
@kstenerud - would you mind giving this a once-over given your experience of implementing on Android? Note - we don't have telemetry in Unity.

kstenerud:
Yup that matches what we do in Android and Cocoa, minus the telemetry.

src/BugsnagUnity/Delivery.cs
rich-bugsnag: Renamed 👍 We could consider a BugsnagConstants class at some point

tomlongridge (approved):

PR #672, opened 2022-11-23
PLAT-9200 Bundle Code not displaying on dashboard

rich-bugsnag:
## Goal

The cocoa bundle version was not being reported unless manually set in the bugsnag config.
If the config value is left blank then it should be automatically reported as the value in the player settings

While investigating we also realised that the android version code had the wrong default value in the settings UI.

## Changeset

- Added a check to only set the cocoa bundle version if a value has been added to the bugsnag config.
- Corrected the default value of the android version code in the settings UI/
## Testing

Added E2E tests for ios and android

### DIFF
--- a/features/fixtures/maze_runner/ProjectSettings/ProjectSettings.asset
+++ b/features/fixtures/maze_runner/ProjectSettings/ProjectSettings.asset
@@ -164,8 +164,8 @@ PlayerSettings:
     iPhone: com.bugsnag.unity.mazerunner
   buildNumber:
     Standalone: 111
-    iOS: 111
-  AndroidBundleVersionCode: 111
+    iOS: 333
+  AndroidBundleVersionCode: 444
   AndroidMinSdkVersion: 16
   AndroidTargetSdkVersion: 0
   AndroidPreferredInstallLocation: 1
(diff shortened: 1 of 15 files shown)


features/fixtures/maze_runner/ProjectSettings/ProjectSettings.asset
```diff
@@ -161,8 +161,11 @@ PlayerSettings:
   applicationIdentifier:
     Android: com.bugsnag.mazerunner
     iOS: com.bugsnag.unity.mazerunner
-  buildNumber: {}
-  AndroidBundleVersionCode: 1
+    iPhone: com.bugsnag.unity.mazerunner
+  buildNumber:
+    Standalone: 111
+    iOS: 111
+  AndroidBundleVersionCode: 111
```
tomlongridge, line 168: Use of the same number leaves the test open to some ambiguity. Can you use and assert different numbers for each of these? I'm not sure what `Standalone` is?

rich-bugsnag: Changed to different values per platform 👍 
Standalone is MacOS

tomlongridge (approved):

PR #675, opened 2022-12-01
Move unity install path on CI and bump version for Windows

joshedney:
## Goal

Move the path of the Unity install on the Windows CI servers
Bump the minor version of the Unity 2021 install

## Testing

Covered by [full ci]

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -602,7 +602,7 @@ steps:
     key: 'windows-2018-fixture'
     depends_on: 'build-artifacts'
     agents:
-      queue: win-general-1-wsl
+      queue: win-general-1
     env:
       UNITY_VERSION: "2018.4.36f1"
     command:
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -604,6 +604,7 @@ steps:
     agents:
       queue: windows-unity-wsl
     env:
+      UNITY_DIR: "/mnt/f/unity/2018.4.36f1"
       UNITY_VERSION: "2018.4.36f1"
```
twometresteve, line 608: Do we really need to provide both?  I wouldn't be too surprised if we did, but it feels messy to have to.  But equally it's not really the remit of this PR to straighten it out - if both are needed then perhaps you could just raise a ticket to review the variables we have and consolidate/rationslise in some way?

.buildkite/pipeline.yml
```diff
@@ -27,8 +27,9 @@ steps:
     agents:
       queue: windows-unity-wsl
     env:
-      UNITY_VERSION: "2018.4.36f1"
-      WSLENV: UNITY_VERSION
+#      UNITY_VERSION: "2018.4.36f1"
```
twometresteve, line 30: Remove if not needed

twometresteve (approved):

.buildkite/pipeline.full.yml
joshedney: Im going to raise a separate ticket for this so that I can allocate some proper time to think of a better solution.

PR #680, opened 2022-12-20
PLAT-9060 transfer device and app data as JSON

rich-bugsnag:
## Goal

Native App and Device data was being serialised as the correct types, causing InvalidCastExceptions in callbacks.
This change converts and passes the data as json, reducing the complexity of the code and retaining the original data types. 

## Changeset

Changed the methods `bugsnag_retrieveDeviceData` and `bugsnag_retrieveAppData` to get JSON from the native layer instead of using complex and costly MonoPInvoke callbacks.

## Testing

csharp_config.feature contained some tests that were modified just for Cocoa to pass. These have been removed and now cocoa runs against the same tests as all other platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Fix an issue where Cocoa Device and App data was serialised incorectly causing invalid cast exceptions in callbacks [#680](https://github.com/bugsnag/bugsnag-unity/pull/680)
+
+
+
+
 ## 7.5.0 (2022-11-30)
 
 ### Enhancements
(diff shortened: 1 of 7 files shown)


twometresteve:
> Native App and Device data was being serialised as the correct types, causing InvalidCastExceptions in callbacks. 

Have we now got one or more e2e tests that would detect a problem like this?

rich-bugsnag:
> > Native App and Device data was being serialised as the correct types, causing InvalidCastExceptions in callbacks.
> 
> Have we now got one or more e2e tests that would detect a problem like this?

Added getter testing and found a tasty bug! Fixed it and all running well now 👍 thanks!

twometresteve (approved):

PR #681, opened 2022-12-21
PLAT-9130 onsession callbacks not reflecting changes made to user

rich-bugsnag:
## Goal

Changes made to the session user in OnSession callbacks did not change anything in the sent payload.

## Changeset

Fixed a simple bug where the user object passed to the payload was different to the one run through the callback.

## Testing

Existing tests had been skipped pending this fix, they are now reactivated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+
+### Bug fixes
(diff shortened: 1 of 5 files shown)


twometresteve (approved):
LGTM

PR #683, opened 2022-12-21
PLAT-9336 Handle errors in delivery

rich-bugsnag:
## Goal

Discard a payload if an error occurs during delivery that is outside the retry rules set out in the notifier spec.
This is important so that no events get stuck in a retry loop.

## Changeset

- Extended the PushToServer coroutine to handle network errors and any exceptions raised during serialisation of the body.

## Testing

Existing E2E tests were enhanced to use the recently added `I set the HTTP status code for the next requests to` step

### DIFF
--- a/features/csharp/csharp_persistence.feature
+++ b/features/csharp/csharp_persistence.feature
@@ -6,7 +6,9 @@ Feature: Unity Persistence
   @skip_windows @skip_macos @skip_webgl #pending PLAT-8632
   Scenario: Receive a persisted session
     When I run the game in the "PersistSession" state
+    And I set the HTTP status code for the next requests to "408"
     And I wait for requests to fail
+    And I discard the oldest session
     And I close the Unity app
     And On Mobile I relaunch the app
     And I run the game in the "PersistSessionReport" state
@@ -19,7 +21,9 @@ Feature: Unity Persistence
 
   Scenario: Receive a persisted event
     When I run the game in the "PersistEvent" state
+    And I set the HTTP status code for the next requests to "408"
     And I wait for requests to fail
+    And I discard the oldest error
     And I close the Unity app
     And On Mobile I relaunch the app
     And I run the game in the "PersistEventReport" state
@@ -33,7 +37,9 @@ Feature: Unity Persistence
 
   Scenario: Receive a persisted event with on send callback
     When I run the game in the "PersistEvent" state
+    And I set the HTTP status code for the next requests to "408"
     And I wait for requests to fail
+    And I discard the oldest error
     And I close the Unity app
     And On Mobile I relaunch the app
     And I run the game in the "PersistEventReportCallback" state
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* Added checks in delivery to ensure payloads cann't get stuck in a retry loop. [#683](https://github.com/bugsnag/bugsnag-unity/pull/683)
```
twometresteve, line 7:
```suggestion
* Added checks in delivery to ensure payloads can't get stuck in a retry loop. [#683](https://github.com/bugsnag/bugsnag-unity/pull/683)
```

features/csharp/csharp_persistence.feature
```diff
@@ -6,7 +6,9 @@ Feature: Unity Persistence
   @skip_windows @skip_macos @skip_webgl #pending PLAT-8632
   Scenario: Receive a persisted session
     When I run the game in the "PersistSession" state
+    And I set the HTTP status code for the next requests to "408"
```
twometresteve, line 9: I think this could be racy and I would switch their order.

features/csharp/csharp_persistence.feature
```diff
@@ -6,7 +6,9 @@ Feature: Unity Persistence
...
     And I wait for requests to fail
+    And I discard the oldest session
```
twometresteve, line 11: I would add a `I wait to receive an error` before discarding the oldest session, but after `I wait for requests to fail`.

It would just add some confidence that the notifier really did try to make a request - and avoid a potentially confusing/messy failure when trying to discard something from an empty list (if one isn't sent).

You might find the `I wait for requests to fail` is no longer needed (and would be great to remove), but as discussed IRL a ticket to look at that separately would be fine.

src/BugsnagUnity/Delivery.cs
```diff
@@ -175,14 +204,24 @@ IEnumerator PushToServer(IPayload payload)
...
                     _payloadManager.SendPayloadFailed(payload);
                 }
+                else if (req.isNetworkError)
+                {
+                    // unity detected a non http related error, remove payload from cache and pending payloads
+                    _payloadManager.RemovePayload(payload);
+                }
+                else
+                {
+                    // sending failed with an uacceptable status code, remove payload from cache and pending payloads
```
twometresteve, line 222:
```suggestion
                    // sending failed with an unacceptable status code, remove payload from cache and pending payloads
```

yousif-bugsnag (approved):
LGTM

src/BugsnagUnity/Delivery.cs
```diff
@@ -175,14 +204,24 @@ IEnumerator PushToServer(IPayload payload)
...
+                    _payloadManager.RemovePayload(payload);
                     StartDeliveringCachedPayloads();
                 }
-                else if (req.isNetworkError || code == 0 || code == 408 || code == 429 || code >= 500)
+                else if ( code == 0 || code == 408 || code == 429 || code >= 500)
                 {
                     // sending failed with no network or retryable error, cache payload to disk
                     _payloadManager.SendPayloadFailed(payload);
                 }
+                else if (req.isNetworkError)
```
twometresteve, line 215: This does just the same as the final `else` - do you really want a specific `else if` for it?

rich-bugsnag: I'm a professional programmer, i swear!

features/csharp/csharp_persistence.feature
rich-bugsnag: I decided to do this properly as part of this PR.
I have done everything suggested and i changed `I wait for requests to fail` to `I wait for requests to persist` which is just a 1 second sleep, i think that should be enough.

twometresteve (approved):

PR #700, opened 2023-02-27
PLAT-9092 missing stacktraces in android notify events

richardelms:
## Goal

If a C# exception is not thrown then it contains no stacktrace. 

When Bugsnag.Notify(System.Exception) is called with an unthrown exception, we generate a backup stacktrace using `System.Diagnostics.StackTrace().GetFrames();`

Previously, when generating this trace we stripped the top 3 frames so that internal frames from the notifier would not be reported. 

The pipeline now automatically marks any frame from our notifier as out of project, and in some cases we were stripping frames that should not be stripped due to compiler optimisations.

In this PR we remove the stripping and allow the pipeline to decide what is and isn't in project.

## Changes to grouping

I was initially concerned that this change would break grouping for existing events but i have tested and this change will only break grouping for events where there are no InProject frames and Notify is called with an unthrown exception which is acceptable.

## Changeset

- Removed hardcoded frame stripping.
- Removed unecessary Notify overloads in Client.cs and Bugsnag.cs

## Testing

- Re-enabled existing Android tests that were disabled due to this issue.
- Existing tests should cover this change
- Added overload check to BugsnagUnity.Tests

### DIFF
--- a/features/fixtures/maze_runner/Assets/Scripts/Main.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Main.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections;
 using System.IO;
 using System.Runtime.InteropServices;
-using BugsnagUnity;
 
 [Serializable]
 public class Command
--- a/features/fixtures/maze_runner/Assets/Scripts/Main.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Main.cs
@@ -139,35 +139,6 @@ public class Main : MonoBehaviour
 #endif
     }
 
-
-
-    // this method never runs and is just used to check that no notify overides are accidentally broken during refactoring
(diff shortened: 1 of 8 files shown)


features/fixtures/maze_runner/Assets/Scripts/Main.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections;
 using System.IO;
 using System.Runtime.InteropServices;
+using BugsnagUnity;
```
tomlongridge, line 7: nit: unused import now

tomlongridge (approved):

features/fixtures/maze_runner/Assets/Scripts/Main.cs
richardelms: removed 👍

PR #701, opened 2023-03-03
updated ui branding

clr182:
## Goal
Update from old branding (Bugsnag) to new branding (BugSnag)

## Changeset

Ui Elements were updated to keep inline with BugSnags new branding

Before:
image

image

image

After:
image

image

image

## Testing

Tested changes by importing package into a new unity project. tested UPM changes with same method

### DIFF
--- a/upm-tools/AssemblyDefinitions/Bugsnag.asmdef
+++ b/upm-tools/AssemblyDefinitions/Bugsnag.asmdef
@@ -1,3 +1,3 @@
 ﻿{
-	"name": "Bugsnag"
+	"name": "BugSnag"
 }
--- a/upm-tools/AssemblyDefinitions/BugsnagEditor.asmdef
+++ b/upm-tools/AssemblyDefinitions/BugsnagEditor.asmdef
@@ -1,7 +1,7 @@
 {
-    "name": "BugsnagEditor",
+    "name": "BugSnagEditor",
     "references": [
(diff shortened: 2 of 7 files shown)


upm-tools/AssemblyDefinitions/Bugsnag.asmdef
```diff
@@ -1,3 +1,3 @@
 ﻿{
-	"name": "Bugsnag"
```
richardelms, line 2: This text will never be seen by a user and the actual filename of the assembly is not changing, so please revert this change and leave as it was

upm-tools/AssemblyDefinitions/BugsnagEditor.asmdef
```diff
@@ -1,7 +1,7 @@
 {
-    "name": "BugsnagEditor",
```
richardelms, line 2: This text will never be seen by a user and the actual filename of the assembly is not changing, so please revert this change and leave as it was

richardelms (changes requested):

upm-tools/AssemblyDefinitions/BugsnagEditor.asmdef
richardelms: @clr182 boop

JamesLindsay0 (approved):

PR #706, opened 2023-04-18
PLAT-9373 maze host config getter

richardelms:
## Goal

Get a config file pushed to Android and iOS devices by mazerunener.
Then set the mazerunner host address with the value contained within.

For more details see here: https://github.com/bugsnag/bugsnag-android/pull/1796

## Changeset

- On Android and iOS, check for the existence of the fixture config file, try 5 times over a perioud of 5 seconds. if not found then use the existing default values.

## Testing

Covered in existing ci tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'xcodeproj'
 
 unless Gem.win_platform?
   # Use official Maze Runner release
-  gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v7.26.0'
+  gem 'bugsnag-maze-runner', '~>7.0'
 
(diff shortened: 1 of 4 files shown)


twometresteve:
Looks like it's working on Android, but the device logs for iOS suggest that it's not finding the file and only working because it falls back to BrowserStack.  Also needs the e2e tests updating to account for line number changes.

twometresteve:
> Looks like it's working on Android, but the device logs for iOS suggest that it's not finding the file and only working because it falls back to BrowserStack. Also needs the e2e tests updating to account for line number changes.

No, I'm lying.  It's just that the file wasn't pushed there at first - apologies.

twometresteve (approved):

PR #716, opened 2023-05-26
PLAT-10165 fixture logging improvements

richardelms:
## Goal

- Improve Logging so that it's easier to debug flakes
- Allow longer for mazerunner to insert config file
- Pull simplified log file from device as artifact


twometresteve (approved):
Just a minor typo

PR #717, opened 2023-06-08
PLAT-10216 prevent JNI call for session and event callbacks if not used

richardelms:
## Goal

prevent JNI call for session and event callbacks if not used.

If there are no C# callbacks then we shouldn't register a native Android callback. This will work around a timing issue for most people as session callbacks aren't widely used.

## Changeset

- Only add on session and on send error callbacks if they are in the c# config
- If addOnSession is called after start, then make the JNI call and register a native callback

## Testing

Manually tested and covered by existing e2e tests

### DIFF
--- a/src/BugsnagUnity/INativeClient.cs
+++ b/src/BugsnagUnity/INativeClient.cs
@@ -104,5 +104,7 @@ namespace BugsnagUnity
 
         bool ShouldAttemptDelivery();
 
+        void AddOnSession();
+
     }
 }
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -64,12 +64,15 @@ namespace BugsnagUnity
 
 
 
-
+        private Configuration _configuration;
 
         private bool CanRunOnBackgroundThread;
 
         private static bool Unity2019OrNewer;
         private Thread MainThread;
+
+        private bool _sessionCallbackSet;
+        private bool _onSendCallbackSet;
 
         private class OnSessionCallback : AndroidJavaProxy
         {
@@ -135,6 +138,7 @@ namespace BugsnagUnity
 
         public NativeInterface(Configuration cfg)
         {
+            _configuration = cfg;
             AndroidJavaObject config = CreateNativeConfig(cfg);
             Unity2019OrNewer = IsUnity2019OrNewer();
             MainThread = Thread.CurrentThread;
@@ -331,9 +335,16 @@ namespace BugsnagUnity
             }
 
             //Register for callbacks
-            obj.Call("addOnSession", new OnSessionCallback(config));
-            obj.Call("addOnSend", new OnSendErrorCallback(config));
-
+            if (config.GetOnSessionCallbacks() != null && config.GetOnSessionCallbacks().Count > 0)
+            {
+                obj.Call("addOnSession", new OnSessionCallback(config));
+                _sessionCallbackSet = true;
+            }
+            if (config.GetOnSendErrorCallbacks() != null && config.GetOnSendErrorCallbacks().Count > 0)
+            {
+                obj.Call("addOnSend", new OnSendErrorCallback(config));
+                _onSendCallbackSet = true;
+            }
 
             // set endpoints
             var notify = config.Endpoints.Notify.ToString();
@@ -1198,6 +1209,23 @@ namespace BugsnagUnity
         public void ClearFeatureFlags()
         {
             AndroidJNI.CallVoidMethod(GetClientRef(), ClearFeatureFlagsMethod, null);
+        }
+
+        public void AddOnSession()
+        {
+            // This was added to android only because we suspect session callbacks might be causing ANRs.
+            // Not many users use session callbacks, so we are only subscribing to the native side if necessary 
+            if (_sessionCallbackSet || _configuration == null)
+            {
+                return;
+            }
+
+            var addOnSessionmethodId = AndroidJNI.GetMethodID(ClientClass, "addOnSession", "(Lcom/bugsnag/android/OnSessionCallback;)V");
+            var callback = new OnSessionCallback(_configuration);
+            object[] args = new object[] { callback };
+            jvalue[] jargs = AndroidJNIHelper.CreateJNIArgArray(args);
+            AndroidJNI.CallVoidMethod(GetClientRef(), addOnSessionmethodId, jargs);
+            _sessionCallbackSet = true;
         }
 
     }
--- a/src/BugsnagUnity/Native/Cocoa/NativeClient.cs
+++ b/src/BugsnagUnity/Native/Cocoa/NativeClient.cs
@@ -449,5 +449,9 @@ namespace BugsnagUnity
         {
             return true;
         }
+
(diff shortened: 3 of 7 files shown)


src/BugsnagUnity/INativeClient.cs
```diff
@@ -104,5 +104,7 @@ interface INativeClient : IFeatureFlagStore
 
         bool ShouldAttemptDelivery();
 
+        void AddOnSession();
```
tomlongridge, line 107: Can we call it something else to avoid confusion with adding C# callbacks:

```suggestion
        void registerForOnSessionCallbacks();
```

src/BugsnagUnity/Native/Cocoa/NativeClient.cs
```diff
@@ -449,5 +449,9 @@ public bool ShouldAttemptDelivery()
         {
             return true;
         }
+
+        public void AddOnSession()
+        {
```
tomlongridge, line 454: I think for completeness we should implement the Cocoa side too as this has the equivalent `bugsnag_registerForSessionCallbacks` method.

src/BugsnagUnity/Native/Fallback/NativeClient.cs
```diff
@@ -164,5 +164,9 @@ public bool ShouldAttemptDelivery()
         {
             return true;
         }
+
+        public void AddOnSession()
+        {
```
tomlongridge, line 169: I'd prefer to see a comment in here to show it's intentionally unimplemented rather than something you hadn't got around to:

```
// Not required
```

src/BugsnagUnity/INativeClient.cs
richardelms: done 👍

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -64,13 +64,16 @@ class NativeInterface
...
-
+        private Configuration _configuration;
 
         private bool CanRunOnBackgroundThread;
 
         private static bool Unity2019OrNewer;
         private Thread MainThread;
 
+        private bool _sessionCallbackSet;
+        private bool _onSendCallbackSet;
```
tomlongridge, line 75: I don't believe this is required?

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -1200,5 +1211,22 @@ public void ClearFeatureFlags()
             AndroidJNI.CallVoidMethod(GetClientRef(), ClearFeatureFlagsMethod, null);
         }
 
+        public void RegisterForOnSessionCallbacks()
+        {
+            // This was added to android only because we suspect session callbacks might be causing ANRs.
+            // Not many users use session callbacks, so we are only subscribing to the native side if necessary
```
tomlongridge, line 1217: Can you add this to the `INativeClient` definition – although I think the lasting justification should be that it's there as an optimisation to avoid unnecessary calls to the native layer if not required. The ANR driver should result in a separate fix.

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -64,13 +64,16 @@ class NativeInterface
...
 
-
+        private Configuration _configuration;
 
         private bool CanRunOnBackgroundThread;
 
         private static bool Unity2019OrNewer;
         private Thread MainThread;
 
+        private bool _sessionCallbackSet;
```
tomlongridge, line 74: Picky as I know this just internal naming, but `_registeredForSessionCallbacks` used in the Cocoa implementation sounds better to me.

src/BugsnagUnity/Native/Android/NativeInterface.cs
richardelms: oops

src/BugsnagUnity/INativeClient.cs
tomlongridge, line 1: I've taken the liberty of making this more generic.

jchowdown:
Hi, I'm investigating why our bugsnag (7.5.2) is crashing inside AddMetadata. Our callstack looks like this:
```

0  libc.so +0x8d394        abort
1  libart.so +0x6f3658     art::Runtime::Abort(char const*)
2  libbase.so +0x16ea4     android::base::SetAborter(std::__1::function<void (char const*)>&&)::$_3::__invoke(char const*)
3  libbase.so +0x1644c     android::base::LogMessage::~LogMessage()
4  libart.so +0x3a2ab4     art::JavaVMExt::AddGlobalRef(art::Thread*, art::ObjPtr<art::mirror::Object>)
5  libart.so +0x5e3594     art::JNI<false>::NewGlobalRef(_JNIEnv*, _jobject*)
6  libunity.so +0x123a28   AndroidJNIBindingsHelpers::NewGlobalRef(void*) (libunity)
7  libil2cpp.so +0x511ad44 GlobalJavaObjectRef__ctor_mFE5679D1B51F51CBF11721773C0D767286AC22E8 (libil2cpp)
8  libil2cpp.so +0x511e684 AndroidJavaObject__ctor_m0CEE7D570807333CE2C193A82AB3AB8D4F873A6B (libil2cpp)
9  libil2cpp.so +0x511d2c8 AndroidJavaObject_AndroidJavaObjectDeleteLocalRef_m2ECEEAF6389ABB9D6B963B8A98568ECD9413DF3C (libil2cpp)
10 libil2cpp.so +0x2d95064 AndroidJavaObject__Call_TisRuntimeObject_m6B809BA99521C27E9641C9A6C5C46A3385ED7B14_gshared (libil2cpp)
11 libil2cpp.so +0x5122348 _AndroidJNIHelper_GetSignature_m414A7A6B98FB5565075E4C51F22C482F2BCEEF5F (libil2cpp)
12 libil2cpp.so +0x48d20d0 _AndroidJNIHelper_GetSignature_TisRuntimeObject_m7CC50A493D0C6897ECB2692A7CEFB3A2CD3655A5_gshared (libil2cpp)
13 libil2cpp.so +0x48d1830 _AndroidJNIHelper_GetMethodID_TisRuntimeObject_m7BB44CFD4D80045298CF68828C0FA8F3E1FFE51C_gshared (libil2cpp)
14 libil2cpp.so +0x2d948ec AndroidJavaObject__Call_TisRuntimeObject_m6B809BA99521C27E9641C9A6C5C46A3385ED7B14_gshared (libil2cpp)
15 libil2cpp.so +0x4b4a570 NativeInterface_DictionaryToJavaMap_m1693BD0E35E76B7E299C7695C0774C2F479938C7 (libil2cpp)
16 libil2cpp.so +0x4b43190 NativeInterface_AddMetadata_m4CC3DB5C41CE2198A9BB0F63B4BBF1B7A20EC781 (libil2cpp)
17 libil2cpp.so +0x2c71ba0 Metadata_AddMetadata_m3B9BBF20CF9D4A232ECA182701FC71009EBA2764 (libil2cpp)
18 libil2cpp.so +0x2c375b8 OurCSharpCode_LogSomeStateAsMetadata_m212DF01A6C8D7B2D5C2623669BC31EF5AEB23714 (libil2cpp)
```

I was wondering if this fix might address our issue (and if so we'll update)?

PR #718, opened 2023-06-09
prevent native calls for session and event callbacks if not used

richardelms:
## Goal

prevent JNI call for session and event callbacks if not used.

If there are no C# callbacks then we shouldn't register a native Android callback. This will work around a timing issue for most people as session callbacks aren't widely used.

## Changeset

- Only add on session and on send error callbacks if they are in the c# config
- If addOnSession is called after start, then make the JNI call and register a native callback

## Testing

Manually tested and covered by existing e2e tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem 'xcpretty'
 gem 'xcodeproj'
 
 # Use official Maze Runner release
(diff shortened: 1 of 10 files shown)


tomlongridge (approved):
LGTM - pending the CI discussions

richardelms:
WebGL CI failing because of backporting env issues, decided to skip instead of fix because of the time required and the changes having nothing to do with web gl in any way

PR #720, opened 2023-06-13
Release v6.3.3

richardelms:
## Bug Fixes

- Prevent callback to Unity from native layers for session and event callbacks if not used to avoid potential ANRs and improve performance. [#717](https://github.com/bugsnag/bugsnag-unity/pull/717)


richardelms:
@tomlongridge Note: we expect webgl to fail on this branch

PR #721, opened 2023-06-16
new team id [full ci]

richardelms:

### DIFF
--- a/features/fixtures/NativeCrashy.xcodeproj/project.pbxproj
+++ b/features/fixtures/NativeCrashy.xcodeproj/project.pbxproj
@@ -272,7 +272,7 @@
 			buildSettings = {
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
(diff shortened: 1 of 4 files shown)


features/ios/ios_native_errors.feature
```diff
@@ -22,6 +22,7 @@ Feature: iOS Native Errors
     And I discard the oldest error
     And the exception "message" equals "Last Run Info Correct"
 
+  @skip_unity_2018
```
twometresteve, line 25: What happened here?  How did changing the team Id make it necessary to skip 2 tests?

twometresteve (approved):
Changes look fine, just concerned why we need to skip two tests now.  Do we need to change something with the new account?

features/ios/ios_native_errors.feature
twometresteve: Oh, I see now - that seems fine to skip it, just bear in mind the need to test the rest of the skipped scenario if we happen to change this area of functionality (seem unlikely).

PR #725, opened 2023-07-03
PLAT-10446 remove unnecessary datetime.now usage

richardelms:
## Goal

It appears that our notifier calling DateTimeOffset.Now is causing some app hangs when down in the Unity framework C++ code. Our hypothesis is that Unity is calling it in a non-thread-safe way or the lock contention is so high that it’s causing the hang.

In fact, for use case shown in the stacktrace we don’t need the actual time and just a timer will suffice. So in the short term we will amend this, and review any other uses of the method.

## Design

In the places mentioned below, DateTime was used to mark a point in time at which the log type rate checks should reset. It was also used as an internal property to mark the time at which a log message was created. All rate limiting is calculated in seconds and so we can use a simple Stopwatch in the client class to keep track of elapsed seconds.

## Changeset
- Added simple timer BSTime to the Client
- Swapped usage of `DateTimeOffset.Now` for simple float timer in the following places:
`MaximumLogTypeCounter` 
`UniqueLogThrottle`
`UnityLogMessage` 

## Testing

Covered by existing tests

### DIFF
--- a/src/BugsnagUnity/Client.cs
+++ b/src/BugsnagUnity/Client.cs
@@ -91,8 +91,8 @@ namespace BugsnagUnity
                 {
                     var unityLogMessage = new UnityLogMessage(exception);
                     var shouldSend = Error.ShouldSend(exception)
-                        && _client._uniqueCounter.ShouldSend(unityLogMessage)
-                        && _client._logTypeCounter.ShouldSend(unityLogMessage);
+                      && _client._uniqueCounter.ShouldSend(unityLogMessage)
+                      && _client._logTypeCounter.ShouldSend(unityLogMessage);
                     if (shouldSend)
                     {
                         var handledState = _config.ReportExceptionLogsAsHandled ? HandledState.ForLoggedException() : HandledState.ForUnhandledException();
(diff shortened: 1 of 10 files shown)


src/BugsnagUnity/Client.cs
```diff
@@ -55,6 +55,21 @@ internal class Client : IClient
...
+            get
+            {
+                if (_bsTimeStopwatch == null)
+                {
+                    _bsTimeStopwatch = new Stopwatch();
+                    _bsTimeStopwatch.Start();
+                }
+                return _bsTimeStopwatch.Elapsed.TotalSeconds;
+            }
+        }
```
lemnik, line 71: Does it possibly make sense to move this into its own class (especially since it's `static`)?

I'm also a bit unsure of the unprotected lazy-init. It looks like we might require some protection around `_bsTimeStopwatch` to be sure that we don't accidentally create more than one. 🤔 

This could either be a double-checked lock:

```
if (_bsTimeStopwatch == null)
{
    lock(lockObject)
    {
        if (_bsTimeStopwatch == null) {
            // init stopwatch here
```

or we could use an atomic reference, but that would require some additional overhead every time we referenced `_bsTimeStopwatch`.

richardelms: good thinking, done 👍

lemnik (approved):
LGTM

PR #727, opened 2023-07-05
Move macOS, Windows and WebGL tests to full pipeline

twometresteve:
## Goal

Move macOS, Windows and WebGL tests for Unity 2020 to full pipeline.

## Design

Aside from also bumping the patch version for Unity 2021, this is a straight cut and paste of Buildkite steps.

## Testing

Covered by triggering a full build after first performing a basic build.


richardelms (approved):
thanks!

PR #731, opened 2023-07-31
PLAT-10671 handle and report deserialisation errors

richardelms:
## Goal

Handle exceptions thrown during deserialisation of a cached events and sessions and log the problematic json so we can debug the issue.

## Changeset

- Added exception handling and logging in Delivery.

## Testing

Not explicitly tested, but existing tests will confirm that it has not broken anything.


SmartbearYing (approved):
LGTM 👍

PR #738, opened 2023-08-17
PLAT-10762 fix android breadcrumb memory leak

richardelms:
## Goal

Some AndroidJavaObject instances were not properly Disposed when passing breadcrumbs through to the native android layer.

## Changeset

- Added a disposable container class that tracks all created objects and disposes of them after use.

## Testing

Created an example app that, when using the latest release, always crashed within 2 seconds of launching due to the leak.
Then rebuilt the app with a new build containing the fix and left it running with a profiler attached for 30 mins. It did not crash and memory usage remaind constant.

### DIFF
--- a/src/BugsnagUnity/Native/Android/NativeInterface.cs
+++ b/src/BugsnagUnity/Native/Android/NativeInterface.cs
@@ -9,6 +9,27 @@ using System.Text;
 
 namespace BugsnagUnity
 {
+    internal class DisposableContainer
+    {
+        private List<IDisposable> _disposables = new List<IDisposable>();
+
+        public void Add(IDisposable disposable)
+        {
+            _disposables.Add(disposable);
+        }
+
+        public void Dispose()
+        {
+            for (int i = _disposables.Count - 1; i >= 0; i--)
+            {
+                if (_disposables[i] != null)
+                {
+                    _disposables[i].Dispose();
+                }
+            }
+        }
+    }
+
     class NativeInterface
     {
         private IntPtr BugsnagNativeInterface;
@@ -658,19 +679,25 @@ namespace BugsnagUnity
             {
                 return;
             }
+            var disposableContainer = new DisposableContainer();
             CallNativeVoidMethod("addMetadata", "(Ljava/lang/String;Ljava/util/Map;)V",
-            new object[] { section, DictionaryToJavaMap(metadata) });           
+            new object[] { section, DictionaryToJavaMap(metadata, ref disposableContainer) });
+            disposableContainer.Dispose();
         }
 
-        internal static AndroidJavaObject DictionaryToJavaMap(IDictionary<string, object> inputDict)
+
+
+        internal static AndroidJavaObject DictionaryToJavaMap(IDictionary<string, object> inputDict, ref DisposableContainer disposable)
         {
             AndroidJavaObject map = new AndroidJavaObject("java.util.HashMap");
+            disposable.Add(map);
             if (inputDict != null)
             {
                 foreach (var entry in inputDict)
                 {
+                    
                     var key = new AndroidJavaObject("java.lang.String", entry.Key);
-
+                    disposable.Add(key);
                     if (entry.Value == null)
                     {
                         map.Call<AndroidJavaObject>("put", key, null);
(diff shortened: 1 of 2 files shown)


src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -9,6 +9,27 @@
 
 namespace BugsnagUnity
 {
+    internal class DisposableContainer
```
lemnik, line 12: Any reason not to make `DisposableContainer` implement `IDisposable` as well?

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -9,6 +9,27 @@
...
+    {
+        private List<IDisposable> _disposables = new List<IDisposable>();
+
+        public void Add(IDisposable disposable)
+        {
+            _disposables.Add(disposable);
+        }
```
lemnik, line 19: Not sure if this is idiomatic, but by adding a little generic this method could return whatever was passed in which would simplify its use:

```suggestion
        public V Add<V>(V disposable) where V : IDisposable
        {
            _disposables.Add(disposable);
            return disposable;
        }
```

which would allow a more concise:
```
var arrayList = disposableContainer.Add(new AndroidJavaObject("java.util.ArrayList"));
```

lemnik (approved):
LGTM - With some ideas

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -658,19 +679,25 @@ public void AddMetadata(string section, IDictionary<string,object> metadata)
...
             CallNativeVoidMethod("addMetadata", "(Ljava/lang/String;Ljava/util/Map;)V",
-            new object[] { section, DictionaryToJavaMap(metadata) });           
+            new object[] { section, DictionaryToJavaMap(metadata, ref disposableContainer) });
+            disposableContainer.Dispose();
         }
 
-        internal static AndroidJavaObject DictionaryToJavaMap(IDictionary<string, object> inputDict)
+
+
+        internal static AndroidJavaObject DictionaryToJavaMap(IDictionary<string, object> inputDict, ref DisposableContainer disposable)
```
gareththackeray, line 690: This doesn't need to be a pass by reference.  It would only need to be so if this method was intending to replace the reference and have that reflected to the caller.

gareththackeray: Also I think `disposables` would be a better parameter name.

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -658,19 +679,25 @@ public void AddMetadata(string section, IDictionary<string,object> metadata)
             {
                 return;
             }
+            var disposableContainer = new DisposableContainer();
```
gareththackeray, line 682: if you make the container `IDisposable` as Jason suggested you'll be able to use `using` here.

src/BugsnagUnity/Native/Android/NativeInterface.cs
richardelms: done 👍

src/BugsnagUnity/Native/Android/NativeInterface.cs
richardelms: I would prefer to keep it as is for readability

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -658,19 +679,24 @@ public void AddMetadata(string section, IDictionary<string,object> metadata)
             {
                 return;
             }
+            using var disposableContainer = new DisposableContainer();
```
gareththackeray, line 682: You need to put the scope in braces or it disposes immediately:

```
using (var disposables = new DisposableContainer()) 
{
  //everything
}
```

gareththackeray: As discussed I'd forgotten about the `using` _declaration_ which uses the parent scope.

jchowdown:
out of curiosity, do we know when this bug was introduced?

PR #740, opened 2023-09-13
added IsStarted method

clr182:
## Goal

Add the public method IsStarted that returns true if the sdk has finished initialising

## Changeset

Added public bool IsStarted() to Bugsnag.cs

## Testing

Added new e2e test to csharp_config.features


richardelms (approved):
Thanks 🙏

PR #750, opened 2023-10-12
Automatically generate PR's when an upstream dependency is updated

Cawllec:
## Goal

When certain notifiers are updated, downstream notifiers need to be updated in a variety of ways in order to test the latest releases.  This PR adds a Github actions workflow and Rake task in order to:

- Generate a pull request updating the upstream submodule whenever it's released, targeting the `next` branch
- Run the full set of tests on the new branch derived from `next`, with the submodule updated

## Design

The workflow, as it is currently implemented, uses the `repository_dispatch` trigger to activate. Upstream, in the `bugsnag-android` or `bugsnag-cocoa` repositories another workflow will be established to trigger this repo their release, triggering this workflow.  The workflow will then set up the environment with the relevant ruby versions and github permissions, then trigger the Rake task `dependencies:update`.

The Rake task leans on `bugsnag/platforms-bumpsnag` to perform certain tasks including:

- Finding and updating the submodule itself
- Parsing the existing changelog and adding a new entry showing the submodule update
- Creating an appropriately named branch and committing the changes with an appropriate message
- Pushing the branch

The workflow will then create a PR with the created branch targeting the `next` branch.  A relevant maintainer will be applied as a reviewer.

## Changeset

This PR is mostly additive, with minor changes made in the Gemfile to install `bumpsnag` in the Github actions test environment.

## Testing

We've been manually testing the workflow changes using a `push` trigger in the new workflow.  This has produced a number of PRs simulating what the output would be, the latest of which [can be found here](https://github.com/bugsnag/bugsnag-unity/pull/749).  Further testing will need to happen once this PR is merged, and a `manual` trigger option will be applied to facilitate that.

Note - the current changelog generated by bumpsnag contains too many line breaks, and will be corrected with a change made there.


.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,56 @@
+name: update-dependencies
+
+on:
+  repository_dispatch:
+    types: [update-dependency]
```
Cawllec, line 5: This will be updated to allow a `manual` trigger both for testing purposes and in production.

Gemfile.lock
```diff
@@ -11,9 +11,9 @@ GEM
       faye-websocket (~> 0.11.0)
       selenium-webdriver (~> 4.2, < 4.6)
     atomos (0.1.3)
-    bugsnag (6.25.2)
+    bugsnag (6.26.0)
```
Cawllec, line 14: All of the changes in the Gemfile.lock will be revoked, as no changes are necessary in this file

Rakefile
```diff
@@ -446,3 +450,43 @@ namespace :test do
...
+
+namespace :dependencies do
+  task :update do
+    target_submodule = ENV['TARGET_SUBMODULE']
+    target_version = ENV['TARGET_VERSION']
+
+    local_info = Bumpsnag.get_git_info
+    target_info = Bumpsnag.get_git_info(target_submodule)
+
+    if target_submodule.nil? || target_version.nil?
```
Cawllec, line 462: This should be moved above the info gathering above.

Rakefile
```diff
@@ -446,3 +450,43 @@ namespace :test do
...
+
+      Bumpsnag.add_changelog_entry(changelog_message, 'Dependencies')
+
+      release_branch = "bumpsnag-#{target_submodule}-#{target_version}"
+
+      commit_message = "[full ci] Update #{target_submodule} to #{target_version}"
+
+      Bumpsnag.change_branch(release_branch, true)
+      Bumpsnag.commit_changes(commit_message, [target_submodule, 'CHANGELOG.md'])
+      Bumpsnag.push_changes(release_branch)
```
Cawllec, line 485: This is quite programmatic, and could be folded into Bumpsnag itself.  This would also allow us to add a bin file to invoke the functions of bumpsnag, allowing us to write scripts outside of Ruby.

Cawllec (commented):
There are a few changes that should be relatively quick to make on this PR.  There are also some larger changes that could be made as Bumpsnag is updated, but shouldn't effect the current progress of this PR.

Rakefile
tomlongridge: I'd certainly be keen to push some of this to Bumpsnag - particularly the changelog message, release branch name and commit message - so they are consistent across other repos. Ideally we'd just pass the submodule names down and it would do the rest?

(Not suggesting we do this before the script is released and in action though.)

Rakefile
```diff
@@ -446,3 +450,43 @@ namespace :test do
...
+      origin_repo = 'https://github.com/bugsnag/bugsnag-unity'
+      target_repo = target_info[:origin]
+
+      changelog_message = "Update #{target_submodule} to [#{target_version}](#{target_repo}/releases/tag/#{target_version}) [##{target_pr}](#{origin_repo}/pull/#{target_pr})"
+
+      Bumpsnag.add_changelog_entry(changelog_message, 'Dependencies')
+
+      release_branch = "bumpsnag-#{target_submodule}-#{target_version}"
+
+      commit_message = "[full ci] Update #{target_submodule} to #{target_version}"
```
tomlongridge, line 481: Does the commit message need to start with `[full_ci]` to be functional or can it go at the end? Ideally we'd bury it at the end so humans don't need to read it.

```suggestion
      commit_message = "[full ci] Update #{target_submodule} to #{target_version}"
```

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+      - run: bundle exec rake dependencies:update
+
+      - name: List current branch name
+        id: current-branch
+        run: echo "branch=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Create pull request
+        if: ${{ steps.current-branch.outputs.branch != 'dep-updater'}}
+        run: >
+         gh pr create -B dep-updater
```
tomlongridge, line 52: Do these need to be changed to `next`?

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+         -H bumpsnag-$TARGET_SUBMODULE-$TARGET_VERSION
+         --title "Update $TARGET_SUBMODULE to version $TARGET_VERSION"
+         --body 'Created by bumpsnag'
+         --reviewer cawllec
```
tomlongridge, line 56: Assume you'll add `rich-bugsnag` in due course?

Rakefile
Cawllec: Agreed - I'll spin this into a separate ticket.

Cawllec: See PLAT-11074

tomlongridge (approved):

PR #755, opened 2023-11-06
Bump Maze Runner version

twometresteve:
## Goal

Prevent certain BugSnag reports being sent to the Maze Runner dashboard.

## Design

This error is being reported in the BugSnag dashboard for Maze Runner, but it's not a problem with Maze Runner as such.
> Test fixture did not GET /command

Changes in later versions of Maze Runner seem to stop this, so it's just a matter of bumping the `Gemlock.lock`.

## Testing

Covered by a full CI run (it's the macOS and WebGL steps that will be affected).


richardelms (approved):
thanks!

PR #756, opened 2023-11-21
Bump Maze Runner version

twometresteve:
## Goal

Fixes an error when trying to run Maze Runner on our ARM-base CI build servers, caused by us having to change how Ruby Gems are installed.

## Testing

Covered by a full CI run.


richardelms (approved):
thanks!

PR #758, opened 2023-11-27
Relax expectation on device language

twometresteve:
## Goal

Relax expectation on device language, as we can't rely on it always being English.

## Testing

Covered by a basic CI run.


richardelms (approved):
good idea, thanks!

PR #760, opened 2023-12-13
Update bugsnag-cocoa to v6.28.0

kstenerud:


tomlongridge (approved):
Changelog not quite right, but can be fixed at release.

PR #766, opened 2024-01-10
Update bugsnag-android v6.5.0

clr182:
## Goal

Update version of bugsnag-android to v6.5.0

## Changeset

- Updated Rakefile to remove call to obsolete `bugsnag-android-ndk` 
- Updated native handling of thread.id, changing from an int to a string
- Changed `redactedKeys` and `discardClasses` to regex patterns when transfering to the native side.

## Testing

Covered by existing e2e tests


twometresteve:
I've changed the PR title to read 6.5.0 rather than 6.1.0 - I assume that was the right thing to do.

YYChen01988 (approved):
LGTM, pending CI

twometresteve (approved):
The change itself looks fine, I just have some questions about testing and docs, if we can have a quick chat IRL when convenient.

PR #773, opened 2024-02-26
PLAT-11641 handle json serialisation errors and remove payloads

richardelms:
### Error handling
Sometimes, while flushing the session and event cache, a serialisation error can occur.

This PR improves the handling of those errors. The new setup is as follows:

- if an error occurs while trying to serialise the json into a dictionary then we silently remove the offending file, as it is most likely caused by an unpreventable IO error.
- if an error occurs while trying to create an event or session from the dictionary then we remove the file and notify the dashboard with the dictionary attached as metadata so that we can try and solve the issue.

ignore errors and delete the payload when turning json string into dictionary
report errors and delete the payload when trying to create an event or session from the dictionary

### File writing

It could be possible the the method File.writeAllText is buffered under the hood. So we are now using a file stream to write data and using the method FileStream.Flush to force it to write all at once.

## Testing

Covered by existing tests

### DIFF
--- a/src/BugsnagUnity/Delivery.cs
+++ b/src/BugsnagUnity/Delivery.cs
@@ -524,25 +524,54 @@ namespace BugsnagUnity
 
                     try
                     {
+                        // if something goes wrong at this stage then we silently discard the file as it's most likley an io error at some point
                         payloadDictionary = ((JsonObject)SimpleJson.DeserializeObject(payloadJson)).GetDictionary();
                     }
                     catch (Exception e)
                     {
-                        Bugsnag.Notify(new Exception("Bugsnag Error. Deserialising a cached payload for delivery failed"),(@event)=>{
-                            @event.AddMetadata("Bugsnag","failedJson",payloadJson);
-                            return true;
-                        });
                         _cacheManager.RemoveCachedEvent(id);
                         continue;
                     }
 
+
+                    // if something goes wrong at this stage then we discard the file and report the error as it might be a bug in the sdk
                     if (isSession)
                     {
-                        Deliver(new SessionReport(_configuration, payloadDictionary));
+                        try
+                        {
+                            var sessionReport = new SessionReport(_configuration, payloadDictionary);
+                            Deliver(sessionReport);
+                        }
+                        catch (Exception e)
+                        {
+                            Bugsnag.Notify(e,(@event)=>
+                            {
+                                @event.Context = "BugSnag Session Cache Error";
+                                @event.AddMetadata("BugsnagSessionData",payloadDictionary);
+                                return true;
+                            });
+                            _cacheManager.RemoveCachedSession(id);
+                            continue;
+                        }
                     }
                     else
                     {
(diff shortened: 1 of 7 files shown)


gareththackeray (approved):

src/BugsnagUnity/Delivery.cs
```diff
@@ -524,21 +524,54 @@ private IEnumerator ProcessCachedItems(Type t)
 
                     try
                     {
+                        // if something goes wrong at this stage then we silently discard the file as it's most likley an io error at some point
```
tomlongridge, line 527: Let's be a bit more specific about this:

```suggestion
                        // if something goes wrong at this stage then we silently discard the file as it's most likely that the file wasn't fully serialised to disk
```

src/BugsnagUnity/Delivery.cs
```diff
@@ -524,21 +524,54 @@ private IEnumerator ProcessCachedItems(Type t)
...
 
+
+                    // if something goes wrong at this stage then we discard the file and report the error as it might be a bug in the sdk
                     if (isSession)
                     {
-                        Deliver(new SessionReport(_configuration, payloadDictionary));
+                        try
+                        {
+                            var sessionReport = new SessionReport(_configuration, payloadDictionary);
+                            Deliver(sessionReport);
```
tomlongridge, line 543: It feels like the `Deliver` call should be outside the catch as it's conceptually not the same thing as we're trying to report.

src/BugsnagUnity/Delivery.cs
```diff
@@ -524,21 +524,54 @@ private IEnumerator ProcessCachedItems(Type t)
...
+                        }
+                        catch (Exception e)
+                        {
+                            Bugsnag.Notify(e,(@event)=>
+                            {
+                                @event.Context = "BugSnag Session Cache Error";
+                                @event.AddMetadata("BugsnagSessionData",payloadDictionary);
```
tomlongridge, line 550: This feels like we're skirting around internal error reports – adding a new tab to a customer's event isn't something I'm aware of us doing elsewhere for notifier issues.

@richardelms – can you take a bit of a look at the Android and Cocoa implementations for Internal Events so you can see how we achieve it and how much work it would be to adopt this mechanism here? I believe it's used for almost exactly this scenario: serialisation/deserialisation issues.

src/BugsnagUnity/Delivery.cs
```diff
@@ -524,21 +524,53 @@ private IEnumerator ProcessCachedItems(Type t)
...
+                        }
+                        else
+                        {
+                            _cacheManager.RemoveCachedEvent(id);
+                        }
                         continue;
                     }
 
+
+                    // if something goes wrong at this stage then we discard the file and report the error as it might be a bug in the sdk
```
tomlongridge, line 544: Probably just drop this comment now? The comment below suggests the future improvement.

tomlongridge (approved):

PR #776, opened 2024-03-01
Grab test app logs after each scenario

Cawllec:
## Goal

Adds functionality to pull custom app logs from the test application after each test run.  These will then be uploaded as part of the maze_output.zip artifact.


richardelms (approved):
thanks!

PR #786, opened 2024-04-09
Drop testing on Unity 2018 and 2019

twometresteve:
## Goal

- Drops testing on Unity 2018 and 2019.
- Updates example app version.
## Design

Testing on iOS/Android for Unity 2021 is moved to the basic pipeline, everything else goes to the full pipeline.

The artifact is now built under Unity 2021.  Initially I tried using 2020, but there seems to be an obscure bug where the export would hang randomly and for no obvious reason.

## Testing

Covered by a full CI run.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -1,6 +1,4 @@
 aliases:
-  - &2018 "2018.4.36f1"
-  - &2019 "2019.4.35f1"
   - &2020 "2020.3.48f1"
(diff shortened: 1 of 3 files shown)


richardelms (commented):
I have tested importing the package built from this project into a 2020 project and it works no problem. BUT i would like to discuss this change further in person before we merge 👍

richardelms (approved):
I have assessed the resulting package and belive that building the notifier with 2021 is the correct approach. Merging this now to get it in release.

PR #797, opened 2024-05-08
PLAT-12109 add bugsnag integity header

richardelms:
## Goal

add bugsnag integity header

## Testing

Covered by existing mazerunner tests


twometresteve:
I'd suggest running the full pipeline to verify that WebGL behaviour is correct and tests pass.

twometresteve (approved):
LGTM

twometresteve:
Actually, should there be a CHANGELOG entry for this?

PR #801, opened 2024-05-16
PLAT-12127 add traceid to events

richardelms:
## NOTE

This PR is part of work connected to the Unity Notifier, the sister PR can be found here: https://github.com/bugsnag/bugsnag-unity-performance/pull/112

Please do not review this PR without also reviewing the other

## Goal

When a C# layer error occurs, get the current span context from the Unity Performance SDK and include it in the event report.

## Changeset

Added PerformanceHelper class to get information from the Performance SDK if it is present.

## Testing

Manually tested
E2E Tests will be added in a later task


lemnik (approved):
LGTM

kstenerud:
I really think that we should version these kinds of cross-library interfaces so that we don't accidentally paint ourselves into a corner...

PR #806, opened 2024-05-30
merge Integration/error correlation into next 

richardelms:
## Goal

work on error correlation is complete, this PR can be merged when we want to release the feature.


src/BugsnagUnity/Payload/Correlation.cs
```diff
@@ -0,0 +1,53 @@
...
+using UnityEngine.Events;
+#nullable enable
+
+namespace BugsnagUnity.Payload
+{
+    public class Correlation : PayloadContainer
+    {
+       
+        private const string TRACE_ID_KEY = "traceid";
+        private const string SPAN_ID_KEY = "spanid";
```
tomlongridge, line 12:
```suggestion
        private const string TRACE_ID_KEY = "traceId";
        private const string SPAN_ID_KEY = "spanId";
```

tomlongridge: @richardelms - would you mind getting this updated and in related tests?

PR #807, opened 2024-06-03
PLAT-12206 treat discardClasses and redactedKeys as Regex

richardelms:
## Goal

Change discardClasses and redactedKeys to regex collections.

Keep the fields in the Bugsnag settings window as string collections so that unity and display them, but convert them to Regex when building the config object.

## Testing

Added Unit Tests and covered by existing E2E tests

### DIFF
--- a/src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
+++ b/src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
@@ -34,7 +34,7 @@ namespace BugsnagUnity
         public bool PersistUser = true;
         public string SessionEndpoint = "https://sessions.bugsnag.com";
         public ThreadSendPolicy SendThreads = ThreadSendPolicy.UnhandledOnly;
-        public string[] RedactedKeys = new string[] { "password" };
+        public string[] RedactedKeys = new string[] { ".*password.*" };
         public string ReleaseStage;
         public bool ReportExceptionLogsAsHandled = true;
         public bool SendLaunchCrashesSynchronously = true;
@@ -79,7 +79,16 @@ namespace BugsnagUnity
 
             config.BreadcrumbLogLevel = GetLogTypeFromLogLevel( BreadcrumbLogLevel );
             config.Context = Context;
-            config.DiscardClasses = DiscardClasses;
+            foreach(string discardedClass in DiscardClasses){
+                try
+                {
+                    config.DiscardClasses.Add(new System.Text.RegularExpressions.Regex(discardedClass));
+                }
+                catch (Exception e)
+                {
+                    Debug.LogError("Regex error adding discard class: " + e.Message);
+                }
+            }
             if (EnabledReleaseStages != null && EnabledReleaseStages.Length > 0)
             {
                 config.EnabledReleaseStages = EnabledReleaseStages;
@@ -98,7 +107,17 @@ namespace BugsnagUnity
             {
                 config.Endpoints = new EndpointConfiguration(NotifyEndpoint, SessionEndpoint);
             }
-            config.RedactedKeys = RedactedKeys;
+            foreach(string key in RedactedKeys)
+            {
+                try
+                {
+                    config.RedactedKeys.Add(new System.Text.RegularExpressions.Regex(key));
+                }
+                catch (Exception e)
+                {
+                    Debug.LogError("Regex error adding redacted key: " + e.Message);
+                }
+            }
             if (string.IsNullOrEmpty(ReleaseStage))
             {
                 config.ReleaseStage = Debug.isDebugBuild ? "development" : "production";
--- a/src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
+++ b/src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
@@ -84,9 +84,9 @@ namespace BugsnagUnity
                 {
                     config.DiscardClasses.Add(new System.Text.RegularExpressions.Regex(discardedClass));
                 }
-                catch (Exception e)
+                catch (ArgumentException e)
                 {
-                    Debug.LogError("Regex error adding discard class: " + e.Message);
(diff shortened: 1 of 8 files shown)


src/BugsnagUnity/Configuration.cs
```diff
@@ -54,15 +55,53 @@ public class Configuration : IMetadataEditor, IFeatureFlagStore
...
                 return false;
             }
-            foreach (var redactedKey in RedactedKeys)
+            if (_redactedKeysRegex == null || _redactedKeysRegex.Count != RedactedKeys.Length)
             {
-                if (key.ToLower() == redactedKey.ToLower())
+                _redactedKeysRegex = new List<Regex>();
+                foreach (var redactedKey in RedactedKeys)
+                {
+                    _redactedKeysRegex.Add(new Regex(redactedKey));
```
tomlongridge, line 71: Do we need some error handling here in case of a illegal regex expression?

src/BugsnagUnity/Configuration.cs
```diff
@@ -54,15 +55,53 @@ public class Configuration : IMetadataEditor, IFeatureFlagStore
 
         internal OrderedDictionary FeatureFlags = new OrderedDictionary();
 
+        private List<Regex> _redactedKeysRegex;
+        public string[] RedactedKeys = new string[] { "password" };
```
tomlongridge, line 59: The old matching was whole word and case insensitive - should we make this the same?

src/Assets/Bugsnag/Scripts/BugsnagSettingsObject.cs
```diff
@@ -79,7 +79,16 @@ public Configuration GetConfig()
...
-            config.DiscardClasses = DiscardClasses;
+            foreach(string discardedClass in DiscardClasses){
+                try
+                {
+                    config.DiscardClasses.Add(new System.Text.RegularExpressions.Regex(discardedClass));
+                }
+                catch (Exception e)
+                {
+                    Debug.LogError("Regex error adding discard class: " + e.Message);
+                }
```
tomlongridge, line 90: Let's make this more specific:

```suggestion
                catch (ArgumentException e)
                {
                    Debug.LogError("Invalid Regex pattern for discard class: " + e.Message);
                }
```

src/BugsnagUnity/Configuration.cs
```diff
@@ -54,15 +55,35 @@ public class Configuration : IMetadataEditor, IFeatureFlagStore
 
         internal OrderedDictionary FeatureFlags = new OrderedDictionary();
 
+
+        public List<Regex> RedactedKeys = new List<Regex>{new Regex(".*password.*",RegexOptions.IgnoreCase)};
         public bool KeyIsRedacted(string key)
         {
-            if (RedactedKeys == null || RedactedKeys.Length == 0)
+            if (RedactedKeys == null || RedactedKeys.Count == 0)
```
tomlongridge, line 62: Seems unneccesary to check length?

```suggestion
            if (RedactedKeys != null)
            {
                foreach (var regex in RedactedKeys)
                {
                    ....
                }
            }
```

src/BugsnagUnity/Native/Android/NativeInterface.cs
```diff
@@ -454,9 +454,14 @@ AndroidJavaObject CreateNativeConfig(Configuration config)
...
-            if (config.DiscardClasses != null && config.DiscardClasses.Length > 0)
+            if (config.DiscardClasses != null && config.DiscardClasses.Count > 0)
             {
-                obj.Call("setDiscardClasses", GetAndroidRegexPatternSetFromArray(config.DiscardClasses));
+                var patternsAsStrings = new string[config.DiscardClasses.Count];
+                foreach (var pattern in config.DiscardClasses)
+                {
+                    patternsAsStrings[config.DiscardClasses.IndexOf(pattern)] = pattern.ToString();
+                }
+                obj.Call("setDiscardClasses", GetAndroidRegexPatternSetFromArray(patternsAsStrings));
```
tomlongridge, line 464: Can this throw an exception in the case where the C# regex pattern is not valid for Java?

tests/BugsnagUnity.Tests/ConfigurationTests.cs
```diff
@@ -82,5 +83,61 @@ public void EndpointValidation()
...
             Assert.IsFalse(config.Endpoints.IsValid);
         }
+
+        [Test]
+        public void RedactedKeysTest()
+        {
+            var config = new Configuration("foo");
+
+            // Default redacted keys
+            Assert.IsTrue(config.KeyIsRedacted("password"));
```
tomlongridge, line 93: I'd add the extra assertion for non-exact match, as we decided this is what we want:

```
            Assert.IsTrue(config.KeyIsRedacted("mypasswordfield"));
```

tests/BugsnagUnity.Tests/ConfigurationTests.cs
```diff
@@ -82,5 +83,61 @@ public void EndpointValidation()
...
+            // Custom redacted keys
+            config2.RedactedKeys.Add(new Regex(".*secret.*", RegexOptions.IgnoreCase));
+            config2.RedactedKeys.Add(new Regex(".*token.*", RegexOptions.IgnoreCase));
+            Assert.IsTrue(config2.KeyIsRedacted("secret"));
+            Assert.IsTrue(config2.KeyIsRedacted("token"));
+            Assert.IsTrue(config2.KeyIsRedacted("password"));
+
+            var config3 = new Configuration("foo");
+
+            // Regex pattern keys
```
tomlongridge, line 107: Not sure I see the difference between these tests and the ones above?

src/BugsnagUnity/Native/Android/NativeInterface.cs
richardelms: added handling 👍

tomlongridge (approved):

PR #816, opened 2024-06-28
PLAT-12363 build unity notifier with dotnet standard 2.0

richardelms:
## Goal

The sdk was being built with legacy frameworks, these frameworks made sense at the time, but now we only support unity 2020 and up, switching to dot net standard 2.0 means we can use all modern IDEs and also means that supporting the eventual change to dotnet 4.0 will be easier.

## Changeset

- Removed unnecessary cake build setup
- Simplified build.sh
- Updated all projects to dot net standard 2.0
- Updated unit tests to run against dot net standard 2.0

## Testing

Full CI Run against existing tests


lemnik (approved):
LGTM

PR #817, opened 2024-07-02
PLAT-12309 Network breadcrumbs

richardelms:
## Goal

Implement network breadcrumbs

## Changeset

- Added BugsnagUnityWebRequest package
- Subscribed to that package on start 
- Automatically log breadcrumbs that are sent via BugsnagUnityWebRequest
- Added LeaveBreadcrumb overload to allow manual request breadcrumbs

## Testing

Added new E2E tests


src/Assets/Bugsnag/Editor/BugsnagAddScriptingSymbol.cs
```diff
@@ -0,0 +1,34 @@
...
+            catch
+            {
+                // Some users might not have a platform installed, in that case ignore the error
+            }
+        }
+    }
+
+    static void SetScriptingSymbol(BuildTargetGroup buildTargetGroup)
+    {
+        var existingSymbols = PlayerSettings.GetScriptingDefineSymbolsForGroup(buildTargetGroup);
```
tomlongridge, line 27: Is this guaranteed to be non-null? `existingSymbols` is used without a check below and will prepend a semi-colon if it's empty.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Dependencies
 
+### Enhancements
+
+- Added automatic and manual newtwork request breadcrumbs. [#817](https://github.com/bugsnag/bugsnag-unity/pull/817)
```
tomlongridge, line 9:
```suggestion
- Added automatic and manual network request breadcrumbs. [#817](https://github.com/bugsnag/bugsnag-unity/pull/817)
```

src/BugsnagUnity/Client.cs
```diff
@@ -639,5 +643,91 @@ public void ClearFeatureFlags()
...
+        }
+
+        public void LeaveNetworkBreadcrumb(UnityWebRequest request, TimeSpan? duration)
+        {
+            if(!Configuration.IsBreadcrumbTypeEnabled(BreadcrumbType.Request))
+            {
+                return;
+            }
+            string statusMessage = request.result == UnityWebRequest.Result.Success ? "succeeded" : "failed";
+            string fullMessage = $"HttpRequest {statusMessage}";
```
tomlongridge, line 691: Have you used "HttpRequest" over "UnityWebRequest" deliberately? The spec says:

> a short identifier for the first- or third-party library used to make the request

src/BugsnagUnity/Client.cs
```diff
@@ -639,5 +643,91 @@ public void ClearFeatureFlags()
...
+        {
+            var queryParams = new Dictionary<string, string>();
+            var querySegments = uri.Query.TrimStart('?').Split('&');
+
+            foreach (var segment in querySegments)
+            {
+                var parts = segment.Split('=');
+                if (parts.Length == 2)
+                {
+                    queryParams[parts[0]] = parts[1];
```
tomlongridge, line 724: We should run these through redacted keys logic

src/BugsnagUnity/Client.cs
```diff
@@ -639,5 +643,91 @@ public void ClearFeatureFlags()
...
+        {
+            if (_requestStartTimes.ContainsKey(request))
+            {
+                _requestStartTimes.Remove(request);
+            }
+        }
+
+        public void LeaveNetworkBreadcrumb(UnityWebRequest request, TimeSpan? duration)
+        {
+            if(!Configuration.IsBreadcrumbTypeEnabled(BreadcrumbType.Request))
```
tomlongridge, line 686: In the spec we say that the config option controls whether automatic breadcrumbs are set and should control whether we listen to events, rather than whether the breadcrumb is left:

> enabledBreadcrumbTypes shall define which system events the notifier will register/listen for in order to collect Breadcrumbs automatically.

That means that if the developer disables the type, we're not listening at all to the events (i.e. there's no overhead) rather than listen but do nothing.

(That probably needs to be applied to navigation breadcrumbs above too.)

richardelms: i understand the idea, but it's our wrapper so if they use the wrapper then we listen to the events.
also, we use the navigation events for more than breadcrumbs, so it doesn't really make sense in these cases i think?

richardelms: Done for the network listeners, not possible to check the navigation config any earlier

tomlongridge (approved):

PR #822, opened 2024-07-04
Remove unnecessary ToArray() when constructing an error

JamesLindsay0:
## Goal

Attempt to improve performance

## Changeset

- Expose the StackTraceLines array as public get instead of requiring the error to enumerate the lines and then collect them into a new array.


richardelms (approved):
Looks good to me

PR #825, opened 2024-07-08
Fix timing issues when Bugsnag started after the focus event

JamesLindsay0:
## Goal

Fix issues where Bugsnag is started after the focus event is received (so the foreground timer was never started).

## Changeset

- Start the foreground stopwatch automatically on startup.
- Use the `Duration` rather than the `ForegroundDuration` to determine if the app has launched as it takes account of the time before Bugsnag was started.

Supersedes #821.


richardelms (approved):
Looks good, but please don't merge yet as this is technically a breaking change.

PR #827, opened 2024-07-15
Bump the NDK version in bugsnag-android-unity to match android

JamesLindsay0:
## Goal

I could not get the plugin to build due to this old ndk version. The bumped version is the one used in the android plugin so it will not break anything.

## Changeset

- Bump the ndk version in bugsnag-android-unity.


richardelms (approved):
nice!

PR #829, opened 2024-07-16
Fix network breadcrumb test

JamesLindsay0:
## Goal

On some platforms other breadcrumbs such as window focus caused the network breadcrumbs test to fail. This is resolved by ignoring breadcrumb types other than request in the test.


richardelms (approved):
great stuff, thanks

PR #831, opened 2024-07-22
Linux support

JamesLindsay0:
## Goal

Add linux support and tests.

## Changeset

- Changes to build scripts to allow linux support
- Documentation on linux setup
- Bump mazerunner version to fix curl version mismatch


richardelms:
Looking great so far, let's try and get the mazerunner stuff in the same PR

src/BugsnagUnity/Native/Linux/NativeClient.cs
twometresteve, line 1: I'd like @richardelms to check this file over when he's back.

twometresteve (commented):
Nice looking change on the whole, good work.

PR #842, opened 2024-10-08
Pipeline refactor

Cawllec:
## Goal

Attempts to make a nice buildkite canvas when all the tests are run, by grouping them logically by the unity version used to build them.  All tests should be retained, and this shouldn't effect any tests capabilities.


richardelms (approved):
love it, thanks!

PR #843, opened 2024-10-25
[PLAT-12879] Add support for getting and searching the currently loaded images

kstenerud:
As part of the il2cpp support, we need to be able to search through the loaded binary images to find a match for instruction addresses that we collect via `il2cpp_native_stack_trace()`.

This PR adds a new class `LoadedImages` that allows you to sync on-demand with what the dynamic loader has loaded using `Refresh()`, and to search for images using `FindImageAtAddress()`.

Xcode's formatter also had a go at one of the objective-c files, so it's a bit noisy.


richardelms (approved):
Thanks!

PR #845, opened 2024-11-01
[PLAT-12880] Add il2cpp build preprocessor

kstenerud:
Add the `--emit-source-mapping` switch when calling il2cpp so that it emits `LineNumberMappings.json` for matching C# code to native stack frames.


richardelms (approved):
👍

PR #848, opened 2024-11-06
Use the dynamic linker's data directly and copy it C# side immediately

kstenerud:
## Goal

The previous native image monitor had a bug in that it was copying the dynamic linker's pointers directly rather than allocating a copy of the data it pointed to. It would then crash when trying to free those pointers if the dynamic loader ever removed an image.

But even without the bug, there would have been a concurrency hole if an image gets deallocated before we lazily fetch the data from the C# side of things.

## Design

The implementation now just treats the C++ side's `NativeLoadedImage` as a glorified struct, and memcpys the data across to C# land. C# then marshals all of the pointed-to data immediately. This 2-step process is protected by the existing mutex for concurrency protection.

A nice effect of this approach is that we don't have to manage memory at all on the C++ side.

## Testing

Tested manually in a macos app and an ios app.

### DIFF
--- a/src/BugsnagUnity.mm
+++ b/src/BugsnagUnity.mm
@@ -6,146 +6,50 @@
 
 extern "C" {
 
-static uint8_t *byteDupOrNull(const uint8_t * const oldBytes, int byteCount) {
-    auto newBytes = new uint8_t[byteCount];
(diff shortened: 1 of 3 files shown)


richardelms (approved):
thanks!

PR #853, opened 2024-11-13
Basic LoadedImages caching. Also renamed StackTrace to PayloadStackTrace so that the name doesn't clash with the system StackTrace.

kstenerud:
Small tweaks to help when building out native image matching:

- LoadedImages now supports caching so that we don't have to transfer data across on every call.
- `StackTrace` renamed to `PayloadStackTrace` so that it doesn't clash with the system `StackTrace`


richardelms (approved):
Thanks 🙏

PR #855, opened 2024-11-18
[PLAT-13024] Send native mapped information in stack traces from system exceptions

kstenerud:
## Goal

On Cocoa devices, get the corresponding native stack trace entries for a system exception, and add the necessary information for the backend to be able to map it back to the CSharp code.

## Design

We now use `il2cpp_native_stack_trace()` to fetch the native frame addresses from the transpiled C++ code, which can then be mapped back to the CSharp code using `LineNumberMappings.json`. This requires knowledge of the binary image the frame address corresponds to, which we can already find using the `LoadedImages` class.

We now send the following per-frame data in Unity stack traces on Cocoa:

```json
{
  "frameAddress": "4395873640",
  "method": "Void Throw()",
  "machoFile": "/private/var/containers/Bundle/Application/D9733826-3F83-4126-9536-BDA4D980DF65/BugsnagUnity.app/Frameworks/UnityFramework.framework/UnityFramework",
  "machoLoadAddress": "4386652160",
  "machoUUID": "4c3afc07-a7af-2036-b53a-18a7828648ef",
  "inProject": true
},
```

I've written up a HOWTO for mapping this data back to the `.cs` file and line number.

## Testing

Added e2e test that verifies the new frame attributes on cocoa.

### DIFF
--- a/Bugsnag/Assets/Bugsnag/Runtime/INativeClient.cs
+++ b/Bugsnag/Assets/Bugsnag/Runtime/INativeClient.cs
@@ -103,13 +103,7 @@ namespace BugsnagUnity
 
         IDictionary<string, object> GetNativeMetadata();
 
-        /// <summary>
-        /// Find the native loaded image that corresponds to a native instruction address
-        /// supplied by il2cpp_native_stack_trace().
-        /// </summary>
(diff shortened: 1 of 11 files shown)


richardelms (approved):
thanks 👍

PR #861, opened 2024-12-03
Release 8.3.0 UAT fixes

richardelms:
## Changeset

- Add `AssemblyInfo.cs` file so that the version is available for the Editor window to display.
- Update `BugsnagAddScriptingSymbol.cs` and IL2CPP BuildPreprocessor to stop warnings in Unity v6.

### DIFF
--- a/Bugsnag/Assets/Bugsnag/Editor/BugsnagAddScriptingSymbol.cs
+++ b/Bugsnag/Assets/Bugsnag/Editor/BugsnagAddScriptingSymbol.cs
@@ -3,47 +3,42 @@ using UnityEditor;
 using UnityEditor.Build;
 #endif
 using UnityEngine;
-[InitializeOnLoad]
-public class BugsnagAddScriptingSymbol : MonoBehaviour
+namespace BugsnagUnity.Editor
 {
-    private const string DEFINE_SYMBOL = "BUGSNAG_UNITY_WEB_REQUEST";
-
-    private static BuildTargetGroup[] _supportedPlatforms = { BuildTargetGroup.Android, BuildTargetGroup.Standalone, BuildTargetGroup.iOS, BuildTargetGroup.WebGL };
-
-    static BugsnagAddScriptingSymbol()
+    [InitializeOnLoad]
+    public class BugsnagAddScriptingSymbol : MonoBehaviour
     {
-        foreach (var target in _supportedPlatforms)
-        {
-            try
-            {
-                SetScriptingSymbol(target);
-            }
-            catch
-            {
-                // Some users might not have a platform installed, in that case ignore the error
-            }
-        }
-    }
+        private const string DEFINE_SYMBOL = "BUGSNAG_UNITY_WEB_REQUEST";
 
-    static void SetScriptingSymbol(BuildTargetGroup buildTargetGroup)
-    {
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -110,7 +110,7 @@ bundle exec maze-runner features/handled_errors.feature
 
 2. Checkout the `next` branch.
 
-3. Set the new version in `CHANGELOG.md`
+3. Set the new version in `CHANGELOG.md` and `Bugsnag/Assets/Bugsnag/Runtime/AssemblyInfo.cs`
```
tomlongridge, line 113: It might be worth adding a Rakefile command to do the version bump now there are >1 places to do this as there's a moderate chance of forgetting.

Suggest at least tasking up the idea if not something we add straightaway.

Bugsnag/Assets/Bugsnag/Editor/BuildPreprocessor.cs
```diff
@@ -9,16 +9,22 @@ public class BuildPreprocessor : IPreprocessBuildWithReport
 
     public void OnPreprocessBuild(BuildReport report)
     {
+        //This will be enabled once IL2CPP source mapping is supported
```
tomlongridge, line 12: I'm not sure I understand the comment - it seems to be enabled now?

Bugsnag/Assets/Bugsnag/Editor/BuildPreprocessor.cs
```diff
@@ -9,16 +9,22 @@ public class BuildPreprocessor : IPreprocessBuildWithReport
...
+        #if UNITY_6000_0_OR_NEWER
```
tomlongridge, line 13: Small suggestion: perhaps it's worth abstracting a `PlayerSettingsCompat` class that contains this conditioning for any access to PlayerSettings? Then we're less likely to forget to use it in future and the code isn't littered with ifdefs.

Bugsnag/Assets/Bugsnag/Editor/BuildPreprocessor.cs
richardelms: was a mistake

Bugsnag/Assets/Bugsnag/Editor/BuildPreprocessor.cs
richardelms: I like this idea, will do it 👍

Bugsnag/Assets/Bugsnag/Editor/BugsnagAddScriptingSymbol.cs
```diff
@@ -1,38 +1,44 @@
 using UnityEditor;
+#if UNITY_6000_0_OR_NEWER
+using UnityEditor.Build;
+#endif
 using UnityEngine;
```
tomlongridge, line 5: Are all these includes (and the if-def) still required now we've abstracted to the compat class for some of it?

richardelms: ah yes, my mistake, well spotted

tomlongridge (approved):

PR #865, opened 2025-01-02
fix for incorrect session handled/unhandled event counts

richardelms:
## Goal

The session handled/unhandled event counts were not updated if the handled status of the event was changed in a callback.

Additionally, events are supposed to have a `unhandledOverridden` property within the severity reason object, this was added.

## Changeset

- Updated `Event.cs` to track changes in the handled state and update session event counts as needed.
- Removed unnecessary property `Event.IsHandled`.
- Added `unhandledOverridden` to severity reason.

## Testing

Added a new E2E callback test

### DIFF
--- a/Bugsnag/Assets/Bugsnag/Runtime/Delivery.cs
+++ b/Bugsnag/Assets/Bugsnag/Runtime/Delivery.cs
@@ -73,7 +73,7 @@ namespace BugsnagUnity
                         {
                             try
                             {
-                                if (!onSendErrorCallback.Invoke(report.Event))
+                                if (!RunEventCallback(onSendErrorCallback, report.Event))
                                 {
                                     return;
                                 }
@@ -96,6 +96,24 @@ namespace BugsnagUnity
             catch
             {
                 // not avaliable in unit tests
+            }
+        }
+
+         private bool RunEventCallback(Func<IEvent,bool> callback, IEvent @event)
+        {
+            try
+            {
+                var initialUnhandledState = @event.Unhandled;
+                var callbackResult = callback.Invoke(@event);
+                if (initialUnhandledState != @event.Unhandled)
+                {
+                    ((Payload.Event)@event).UnhandledOverridden();
+                }
+                return callbackResult;
+            }
+            catch
+            {
+                return true;
             }
         }
 
--- a/Bugsnag/Assets/Bugsnag/Runtime/Delivery.cs
+++ b/Bugsnag/Assets/Bugsnag/Runtime/Delivery.cs
@@ -67,13 +67,15 @@ namespace BugsnagUnity
                 var report = (Report)payload;
                 if (_configuration.GetOnSendErrorCallbacks().Count > 0)
                 {
+                    // save handled state before callbacks so we can check later if it was overidden
+                    var initialUnhandledState = report.Event.Unhandled;
                     lock (_callbackLock)
                     {
                         foreach (var onSendErrorCallback in _configuration.GetOnSendErrorCallbacks())
                         {
                             try
                             {
-                                if (!RunEventCallback(onSendErrorCallback, report.Event))
+                                if (!onSendErrorCallback.Invoke(report.Event))
                                 {
                                     return;
                                 }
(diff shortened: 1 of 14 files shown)


Bugsnag/Assets/Bugsnag/Runtime/Payload/Event.cs
```diff
@@ -198,31 +200,38 @@ internal void AddAndroidProjectPackagesToEvent(string[] packages)
...
+                    }
+                    else
+                    {
+                        Session.Events.IncrementHandledCount();
+                        Session.Events.DecrementUnhandledCount();
+                    }
+                }
+                Add("unhandled", value);
             }
         }
```
tomlongridge, line 231: This feels like an unexpected place to update these counts as the update only needs to be done in a particular scenario - i.e. when a callback has updated the handledness. I think the implementation on other notifiers puts this logic in the code that triggers the callbacks – storing the original value before callbacks and updating the counts iff the value has changed.

Could that be considered here? Perhaps that means we'd need the logic in multiple places?

Bugsnag/Assets/Bugsnag/Runtime/Payload/Event.cs
```diff
@@ -162,6 +162,8 @@ internal void AddAndroidProjectPackagesToEvent(string[] packages)
             _androidProjectPackages = packages;
         }
 
+        private bool _originalUnhandledState;
```
tomlongridge, line 165: We already have a ticket (PLAT-9619) that we're not setting `unhandledOverridden` in the severity reason. I think we should consider setting this as part of this change.

features/fixtures/maze_runner/Assets/Scripts/Scenario.cs
```diff
@@ -157,7 +157,7 @@ public bool SimpleEventCallback(IEvent @event)
         @event.AddMetadata("test1", new Dictionary<string, object> { { "test", "test" } });
         @event.AddMetadata("test2", new Dictionary<string, object> { { "test", "test" } });
         @event.ClearMetadata("test2");
-
+        @event.Unhandled = true;
```
tomlongridge, line 160: I'd prefer to see this tested both ways around to ensure it also works to set unhandled to handled - can we add a further scenario so we test both ways and also without any change?

Bugsnag/Assets/Bugsnag/Runtime/Delivery.cs
```diff
@@ -73,7 +73,7 @@ public void Deliver(IPayload payload)
                         {
                             try
                             {
-                                if (!onSendErrorCallback.Invoke(report.Event))
+                                if (!RunEventCallback(onSendErrorCallback, report.Event))
```
tomlongridge, line 76: In the unlikely event of there being multiple callbacks that switch the handledness back and forwards, we'd still end up with `UnhandledOverridden` being set.

Given this is only called from one place, suggest you inline the logic here and store the original value before the callbacks and check it after all the callbacks.

Bugsnag/Assets/Bugsnag/Runtime/Payload/Session.cs
```diff
@@ -96,32 +89,36 @@ internal Session Copy()
...
+            this.AddToPayload(UNHANDLED_KEY, unhandled);
         }
 
-        internal int Handled => this["handled"];
-        internal int Unhandled => this["unhandled"];
+        internal int Handled => this[HANDLED_KEY];
+        internal int Unhandled => this[UNHANDLED_KEY];
 
-        internal void IncrementHandledCount()
+        internal void UpdateEventCount(bool handled, bool increment)
```
tomlongridge, line 106: It took me a fair while to parse this bit of logic: the two booleans make quite hard to work out what's happening from the callers side.

Perhaps we could refactor slightly to `AdjustEventCount(bool, int)` where the int is the number to increment, and could be negative. I appreciate it's only going to be 1 or -1, but it's easier to tell what the function is doing without going to the implementation. I also wouldn't have the override with the default as it's non-obvious whether handled or unhandled would be the default.

richardelms: ive changed it so that you can now only either call `UpdateUnhandledCount(bool increment)` or `UpdateHandledCount(bool increment)` is that clear enough? passing an int feels wrong to me as we only ever want to deal with +1 or -1

tomlongridge (approved):

PR #867, opened 2025-01-06
enable minification support

richardelms:
## Goal

When minification is enabled some classes from the native android sdk are stripped because calls from the unity C# to the native android SDK are not detected by the minifier and so it doesnt think they are used.

To fix this a pro guard file containing `-keep class com.bugsnag.android.** {*;}` has been added to the bugsnag android unity module.

## Testing

Minification has been enabled in the android e2e fixture so that we detect if anything is stripped.


lemnik (approved):
LGTM

PR #870, opened 2025-01-15
auto symbol upload testing

richardelms:
## Goal

Implement e2e testing of automatic symbol uploads

## Design

The build process has been converted to a mazerunner scenario that expects and then inspects the contents of source maps. The resulting build is then used to run the tests for all scenarios on that platform.

## Changeset

- added new build feature files with scenarios for android and ios.
- updated pipeline files to use the new methods.
- Created a special getter method `BugsnagSettingsObject.LoadBuildTimeSettingsObject` for editing config values at build time.
- pass application identifier for android uploads.
- removed duplicate rake gem dep.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ gem 'rake'
 gem 'xcpretty'
 gem 'xcodeproj'
 gem 'cocoapods'
-gem 'rake'
 gem 'danger'
 
 unless Gem.win_platform?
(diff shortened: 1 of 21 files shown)


Gemfile
```diff
@@ -4,7 +4,6 @@ gem 'rake'
 gem 'xcpretty'
 gem 'xcodeproj'
 gem 'cocoapods'
-gem 'rake'
```
richardelms, line 7: removed duplicate dependency

tomlongridge (approved):

PR #874, opened 2025-01-30
build and test upm package on CI

richardelms:
## Goal

Before this change the UPM packaged released was always generated and import tested locally. Now it's done in the same moment as the Unity Package meaning there should never be a chance for differences between the released artifacts.

## Changeset

- build the upm package in the same command as the Unity Package.
- Upload it in a zip
- Test importing it into the fixture, if anything is wrong then compilation errors should happen and the step will fail.

## Testing

Currently only testing with Unity 2021, I don't think it's necessary to test every unity version as UPM compatibility is very general across the versions we support.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -62,8 +62,6 @@ steps:
     env:
       UNITY_VERSION: *2021
     key: "import-upm-2020"
-    env:
-      UNITY_VERSION: *2020
     plugins:
       artifacts#v1.9.0:
         download:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -13,13 +13,31 @@ steps:
...
       automatic:
         - exit_status: "*"
           limit: 1
 
+  - label: "Test UPM Package Import"
+    depends_on: build_unitypackage
+    timeout_in_minutes: 10
+    env:
+      UNITY_VERSION: *2021
+    key: "import-upm-2020"
```
tomlongridge, line 31: Not sure what the relevance of the key is, but presumably should be 2021?

```suggestion
    key: "import-upm-2021"
```

tomlongridge (approved):
LGTM, just one minor query

PR #884, opened 2025-02-21
add e2e smoke test for edm4u upm package

richardelms:
## Goal

there are no working e2e tests to make sure that our EDM4U package is working as expected.

## Changes

- Update EDM Fixture.
- Include EDM4U package in built artifacts (this is also useful for our release process).

## Testing

Added a smoke test to make sure that the correct android libs are dynamically pulled in and that an error can be sent.
If something goes wrong then the SDK would not be able to init and no errors would be sent.
This is just a smoke test as not notifier code is changed and we only really need to know that the SDK can init.


features/edm4u/edm4u.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Android EDM4U smoke test
+
+  Scenario: Android EDM4U smoke test
+    When On Mobile I relaunch the app
+    And I wait to receive an error
```
twometresteve, line 5: In theory this _could_ be a source of a flake if the app always send an error when launched.  The app is launched by MR before the scenario starts, so there are 2 requests being generated and in principle they could be received in the same 0.1s interval that MR checks for request with.  To guard against that I'd suggest changing as follows:

```suggestion
    When On Mobile I relaunch the app
    And I wait to receive at least 1 error
```

twometresteve (approved):

PR #885, opened 2025-02-24
Android il2cpp stackframes

lemnik:
## Goal
Report the native addresses from il2cpp on Android.

## Changes
Report the `FrameAddress` as returned from `il2cpp_native_stack_trace`, along with the `codeIdentifier` and `file` so that the frames can be identified and symbolicated.

## Testing
Manual testing and a new end-to-end feature based on existing smoke test scenarios.


richardelms (approved):
many thanks!

PR #889, opened 2025-03-03
Remove direct uses of Maze.driver

twometresteve:
## Goal

Remove direct uses of Maze.driver, as it does not handle failed drivers.  

## Design

Maze Runner's new API methods should be used instead - but as it turns out, these steps are not used anyway and so I'm just removing them as a cleanup operation.

## Testing

Covered by a basic CI run.


twometresteve:
Bad change - it only looked like the steps weren't used because I had a file extension filter active on my IDE search.

PR #893, opened 2025-03-12
Il2cpp abstraction & Cocoa

lemnik:
## Goal
Abstract the common il2cpp stacktrace logic into a common class used by both Android & Cocoa to report native stack frames

## Testing
Introduced a new end-to-end scenario for iOS with il2cpp stack frames


richardelms (approved):
thanks!

PR #899, opened 2025-04-08
verify code formatting

richardelms:
## Goal

Add rake commands that format csharp code and also a CI step that verifies code formatting as part of a CI run.


scripts/code_format.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
twometresteve, line 1: I think you want a `-e` on this so that any failing command sends its exit code back to the shell that ran the script.

joshedney (approved):
LGTM!

PR #900, opened 2025-04-11
Detect WebGL test fixture build failures

twometresteve:
## Goal

Detect WebGL test fixture build failures by checking if `index.html` is in the build folder.  

## Design

The build folder is still zipped and attached to the Buildkite job for inspection should a failure be detected.  Returning non-zero means that Buildkite fails the job correctly and it can be retried either automatically or manually.

## Testing

Naturally tested as part of preparing this PR - the Unity build failed twice on macos-14-5, it was detected and the Buildkite job failed as a result.


joshedney (approved):
LGTM!

PR #904, opened 2025-05-05
add compilation platform error type

richardelms:
## Goal

add il2cpp or mono as error type so that the pipeline can attempt to symbolicate IL2CPP errors if possible.

## Testing

Added new E2E tests

### DIFF
--- a/Bugsnag/Assets/Bugsnag/Runtime/Payload/Error.cs
+++ b/Bugsnag/Assets/Bugsnag/Runtime/Payload/Error.cs
@@ -16,6 +16,14 @@ namespace BugsnagUnity.Payload
 
         internal bool IsAndroidJavaException;
 
+#if ENABLE_IL2CPP
+        const string COMPILATION_PLATFORM = "il2cpp"; 
+#elif ENABLE_MONO
+        const string COMPILATION_PLATFORM = "mono"; 
+#else
+        const string COMPILATION_PLATFORM = "unknown"; 
+#endif
+
         private const string ANDROID_JAVA_EXCEPTION_CLASS = "AndroidJavaException";
         private const string ERROR_CLASS_MESSAGE_PATTERN = @"^(?<errorClass>\S+):\s+(?<message>.*)";
         private const string NATIVE_ANDROID_ERROR_CLASS = "java.lang.Error";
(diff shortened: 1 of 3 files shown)


Bugsnag/Assets/Bugsnag/Runtime/Payload/Error.cs
```diff
@@ -76,8 +86,11 @@ public string ErrorMessage
             set => this.AddToPayload(MESSAGE_KEY, value);
         }
 
-        public string Type { get => "Unity"; }
```
tomlongridge, line 79: We document this field ([here](https://docs.bugsnag.com/platforms/unity/customizing-error-reports/#:~:text=of%20the%20stacktrace-,Type,-ErrorType)) – so this is changing the value of a public field. That said, we don't document its potential values and it's hard-coded right now, so any conditioning on this value in a callback won't be doing an awful lot.

I just wanted to flag as I think it's at least worth noting on the PR description in case this does have an unforeseen consequence.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+- Add compilation platform type to error payloads to enable IL2CPP symbolication in future [#904](https://github.com/bugsnag/bugsnag-unity/pull/904)
```
tomlongridge, line 7:
```suggestion
- Add compilation platform as `error.type` to event payloads to enable IL2CPP symbolication in future [#904](https://github.com/bugsnag/bugsnag-unity/pull/904)
```

Bugsnag/Assets/Bugsnag/Runtime/Payload/Error.cs
```diff
@@ -16,6 +16,14 @@ public class Error : Dictionary<string, object>, IError
 
         internal bool IsAndroidJavaException;
 
+#if ENABLE_IL2CPP
+        const string COMPILATION_PLATFORM = "il2cpp"; 
+#elif ENABLE_MONO
+        const string COMPILATION_PLATFORM = "mono";
+#else
+        const string COMPILATION_PLATFORM = "unknown"; 
+#endif
```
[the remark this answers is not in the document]
richardelms: i like this idea

Bugsnag/Assets/Bugsnag/Runtime/Payload/Error.cs
richardelms: Agreed

tomlongridge (approved):

PR #929, opened 2025-09-05
Update unity tests to cover line number mapping files

joshedney:
## Goal

Update the Unity build steps to check if the unity line mapping files are uploaded as part of the auto upload step

## Testing

Covered by CI


richardelms (approved):
excellent, thanks

PR #941, opened 2025-09-19
check for out of bounds in native trace builder on android

richardelms:
This pull request updates the logic in the `ToStackFrames` method to improve safety when constructing stack traces from native addresses and lines. The main change ensures that the method does not attempt to access elements beyond the bounds of either array, preventing potential runtime errors.

Stack trace construction safety:

* The stack trace array length is now set to the minimum of `nativeAddresses.Length` and `lines.Length`, and the loop iterates up to this minimum length to avoid out-of-bounds access.


lemnik (approved):
LGTM

PR #946, opened 2025-10-15
Make test scenario exception messages more distinctive

twometresteve:
## Goal

Make test scenario exception messages more distinctive, so that requests bleeding across scenarios are much easier to identify.

## Testing

Covered by a basic CI run (only CSharp scenarios have been changed, the rest were already distinctive).


richardelms (approved):
Good stuff, thanks!

PR #948, opened 2025-10-16
Add "zero build" CI level

twometresteve:
## Goal

Introduces a level of CI that performs no build or test steps at all, allowing commits to be pushed for backup only without using any CI resources.

## Design

Extends the existing mechanism for triggering a full build after a basic build, to include a trigger for the initial "basic" build from a "zero" build.  

The build levels are now:
- "Zero" - the default, no build steps, allows manual triggering of a basic build.
- "Basic" -  builds the notifier package and runs a subset of tests for Unity 2021, allows manual triggering of a full build.  Automatically triggered if the commit message includes `[basic ci]'.
- "Full" - Run all e2e tests.  Automatically triggered if the commit message includes `[full ci]'.

Branch protection rules have been put in place to ensure that basic builds are run for all PRs into `next` and a full build gets run for PRs into `master`.

## Testing

Tested with a series of commits with and without the special commit messages, also triggering the basic and full build manually.


.buildkite/pipeline.basic.yml
twometresteve, line 1: This file is a straight cut and paste from the original pipeline.yml file, there's no need to review in on a line-by-line basis.

.buildkite/pipeline.full.yml
```diff
@@ -3,21 +3,25 @@ agents:
 
 steps:
   - label: 'Append Unity 2020 Pipeline'
+    depends_on: "build-unitypackage"
```
twometresteve, line 6: These are needs so that we can trigger a full build without waiting for the notifier package to build first.

.buildkite/pipeline.yml
```diff
@@ -1,92 +1,10 @@
-aliases:
-  - &2021 "2021.3.45f1"
-
-agents:
-  queue: macos-15
-
 steps:
 
-  - label: Verify code formatting
```
twometresteve, line 9: As above, this portion was simply moved to the new file.

richardelms (approved):
thanks

PR #949, opened 2025-10-20
Use Maze Runner idempotent commands

twometresteve:
## Goal

Convert the e2e tests to use idempotent Maze Runner commands, together with various other stability improvements.

## Design

The idempotent command pattern works by passing the UUID of the last successfully processed command when making a command for the next.  This means the fixture needs to write the the UUID to disk once received so that it can be read after restarting the app.

In the course of making the change I realised that there were some inherent race conditions relating to how we sent commands to run scenarios and how were were starting and stopping the app.  We were often ending up with two instances of the app running (on macOS/Windows), causing various forms of test flakes.

## Changeset

- Test fixture now uses the `/idem-command` endpoint, passing the last known UUID.
- Mechanism added to persist the UUID of the last command processed, as well as reset it when instructed to by Maze Runner.
- Fixture app is now opened `Before` every scenario and closed `After` each.  If a scenario needs the app to stop/start then this is now handled by explicit Cucumber steps.
- Approach to starting and stopping the app on macOS/Windows, adding a mechanism to wait for the macOS app to actually stop. 

## Testing

Covered by a full CI run.  It's not completely flake free, but should be much better than it has been.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,4 @@ UPMImportProject.sln
 upm-edm4u-package
 upm-edm4u-package.zip
 testResults.xml
+features/fixtures/maze_runner/mazerunner_BackUpThisFolder_ButDontShipItWithYourGame/
--- a/Gemfile
+++ b/Gemfile
@@ -8,10 +8,10 @@ gem 'danger'
 
 unless Gem.win_platform?
   # Use official Maze Runner release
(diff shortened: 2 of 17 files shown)


richardelms (approved):
Excellent, many thanks!

PR #950, opened 2025-10-20
Ensure paths passed to the Bugsnag CLI are wrapped in quotes

joshedney:
## Goal

Ensure that file paths passed to the Bugsnag CLI are wrapped in quotes to handle spaces in file names 

## Testing

Tested manually by setting the name of the project to one with a space before building. Here is the CI run with that [change](https://buildkite.com/bugsnag/bugsnag-unity/builds/6204/steps/canvas)


richardelms (approved):
thanks a bunch!!!

Bugsnag/Assets/Bugsnag/Editor/SymbolUpload/BugsnagCLI.cs
tomlongridge, line 149: Another point that might need quoting?

tomlongridge, line 165: I think this path might need quoting as it can also be derived from `Application.dataPath`

tomlongridge: ... or do it inside `StartProcess`.

tomlongridge: Discussed IRL - this is covered.

PR #952, opened 2025-10-22
Buildkite pipeline tidy-up

twometresteve:
## Goal

The functional change here is to change this artifact upload pattern:
```
              - "maze_output/*/*"
```
into...
```
              - "maze_output/failed/*"
              - "maze_output/maze_output.zip"
```

This makes it easier for us to find the log files for failed scenarios (which we tend to care about most), whilst still allowing us to access the logs for all scenarios if we wish.

## Testing

Covered by a full CI run.


richardelms (approved):
excellent, thanks

PR #961, opened 2026-01-12
Ensure we're using the latest versions of iOS for testing

joshedney:
## Goal
Update the iOS versions we're using for tests

## Testing
Covered by CI


.buildkite/unity.2021.full.yml
```diff
@@ -329,8 +329,8 @@ steps:
               - "features/ios"
               - "--app=features/fixtures/maze_runner/mazerunner_dev_2021.ipa"
               - "--farm=bb"
-              - "--appium-version=1.22"
-              - "--device=IOS_13|IOS_14|IOS_15"
+              - "--appium-version=2.5.0"
```
twometresteve: I'd remove the --appium-version option as BitBar ignores it, unlike BrowserStack.

twometresteve (approved):

PR #982, opened 2026-04-17
Appium version change Unity

SB-dimplea:
## Goal

Bumped --appium-version from 1.22 to 2.19 in the Unity Android Buildkite pipelines.

## Design

Earlier pipeline was using appium version 1.22 so we have changed appium version 2.19 commonly.

## Changeset

Changed Pipeline.yml files to update version of appium from 1.22 to 2.19
Update app package and app activity in respective yml Files.

## Testing

Yes, We have tested manual and automated maze runner test cases BB.


SB-shivamgarg (changes requested):
please look at the ed4u test case on build kite.

SB-shivamgarg (approved):

twometresteve (approved):
LGTM.  My only query is that the PR description mentions the Maze runner config file and it seems that did not need to change in the send, so could be removed from the description.

PR #986, opened 2026-04-28
update contributing guide

richardelms:
update contributing guide


SB-shivamgarg (changes requested):
We implemented [rake bump VERSION=”X.Y.Z”] command for release process. I think we should add this as well.

PR #992, opened 2026-05-21
PLAT-7972 scripting metadata improvement

SB-jigneshR:
## Goal

Improve reporting of scripting metadata when bugsnag start manually.

## Design

The detection logic has been moved into Configuration.cs as field initialisers, using the same #if preprocessor directives that were previously private to BugsnagAutoInit:
Configuration.cs is a source file compiled as part of the developer's Unity project (not a precompiled DLL), the #if directives resolve against the developer's own scripting backend and API compatibility settings at their compile time. This means any new Configuration(apiKey) call automatically carries the correct values, regardless of whether auto-init or manual start is used.
BugsnagAutoInit no longer needs to set these fields explicitly and the methods have been removed from it.

## Changeset

Configuration.cs — ScriptingBackend, DotnetScriptingRuntime, and DotnetApiCompatibility fields now default to auto-detected values via new private Detect* static methods. Fields remain public so developers can still override them if needed.
BugsnagAutoInit.cs — Removed the now-redundant explicit assignments and private FindScriptingBackend(), FindDotnetScriptingRuntime(), FindDotnetApiCompatibility() methods.
ScriptingBackendPopulated.cs — New maze runner scenario that constructs Configuration manually without setting the scripting backend fields, exercising the auto-detection path.
MainScene.unity — Registered ScriptingBackendPopulated as a component on the Config GameObject so the scenario runner can discover it.
csharp_config.feature — New scenario Scripting backend fields are auto-populated for manual start that asserts device.runtimeVersions.unityScriptingBackend, device.runtimeVersions.dotnetScriptingRuntime, and device.runtimeVersions.dotnetApiCompatibility are all non-null in the reported payload.

## Testing

New maze runner scenario ScriptingBackendPopulated verified end-to-end: the scenario deliberately omits manual assignment of the scripting backend fields and confirms all three values are present and non-null in the error payload sent to the notifier API.
Existing csharp_config suite continues to pass unmodified — the Configuration field initialiser change is backward-compatible; any code that was already setting these fields explicitly will simply overwrite the defaults.

### DIFF
--- a/features/fixtures/maze_runner/Assets/Scenes/MainScene.unity
+++ b/features/fixtures/maze_runner/Assets/Scenes/MainScene.unity
@@ -148,6 +148,7 @@ GameObject:
   - component: {fileID: 355371107}
   - component: {fileID: 355371108}
   - component: {fileID: 355371109}
+  - component: {fileID: 355371110}
   m_Layer: 0
   m_Name: Config
   m_TagString: Untagged
@@ -470,6 +471,18 @@ MonoBehaviour:
 
     Main.CUSTOM2
     () (at Assets/Scripts/Main.cs:123)'
+--- !u!114 &355371110
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 355371092}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 7362c3e39d438464492bbb99c0ff80cf, type: 3}
+  m_Name: 
+  m_EditorClassIdentifier: 
 --- !u!1 &372057234
 GameObject:
   m_ObjectHideFlags: 0
--- a/features/fixtures/maze_runner/Assets/Scripts/Scenarios/Csharp/Config/ScriptingBackendPopulated.cs
+++ b/features/fixtures/maze_runner/Assets/Scripts/Scenarios/Csharp/Config/ScriptingBackendPopulated.cs
@@ -1,10 +1,17 @@
+using BugsnagUnity;
+using UnityEngine;
+
 public class ScriptingBackendPopulated : Scenario
 {
     public override void PrepareConfig(string apiKey, string host)
     {
-        // Deliberately do NOT set ScriptingBackend, DotnetScriptingRuntime, or
-        // DotnetApiCompatibility — they should be auto-detected by Configuration.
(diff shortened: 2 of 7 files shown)


features/fixtures/maze_runner/Assets/Scripts/Scenarios/Csharp/Config/ScriptingBackendPopulated.cs
```diff
@@ -0,0 +1,14 @@
+public class ScriptingBackendPopulated : Scenario
+{
+    public override void PrepareConfig(string apiKey, string host)
+    {
+        // Deliberately do NOT set ScriptingBackend, DotnetScriptingRuntime, or
+        // DotnetApiCompatibility — they should be auto-detected by Configuration.
+        base.PrepareConfig(apiKey, host);
+    }
```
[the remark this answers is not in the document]
SB-shivamgarg: Considering the suggestion

features/fixtures/maze_runner/Assets/Scenes/MainScene.unity
```diff
@@ -470,6 +471,18 @@ MonoBehaviour:
...
+--- !u!114 &355371110
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 355371092}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 7362c3e39d438464492bbb99c0ff80cf, type: 3}
```
SB-sourabhk, line 483: do u need to show guid ? how can we avoid showing any 128 bit random string

SB-shivamgarg: In unity guid is work like a file path. Unity references assets by GUID. It's not a secret. Because this a scene file and we created a new test file and attached with this scene so this GUID is showing as new change.

SB-sourabhk: I understood it’s not a secret but can we have only fileID and not the whole object {} mentioning guid.. just a suggestion. I would recommend only fileID to be shown like rest of the keys.

SB-shivamgarg: Unfortunately not. In Unity, fileID is only unique within an asset, while the guid identifies the asset itself. Unity requires both values for script references and automatically serializes them in scene files.

SB-sourabhk (approved):

SB-meiyalaganr (approved):
LGTM

PR #993, opened 2026-06-03
Update .gitignore to include additional files

SB-shivamgarg:
## Goal

Strengthen .gitignore file

## Changeset

Added new items in the .gitignore file and ignore some files based on extension which can expose secrets and sensitive information


.gitignore
```diff
@@ -43,18 +50,11 @@ Bugsnag.meta
 unity/PackageProject/Assets/Bugsnag
 nunit-agent*.log
```
SB-sourabhk: as u already removing all *.log file so is it required ? can be cleared ?

.gitignore
```diff
@@ -1,11 +1,15 @@
...
 local.properties
 temp.unityproject
 bugsnag-osx/bugsnag-osx/Bugsnag.framework
 xcuserdata
-xcodebuild.log
+*.log
+*.so
 build.log
 *.unitypackage
 .vs
```
SB-sourabhk: isn’t this should be *.vs or any specific .vs file

SB-shivamgarg: Done

SB-sourabhk (approved):

SB-jigneshR (approved):

